PR #17, opened 2021-06-22
[BACK-1951] Remove patient profiles from clinics when all permissions are revoked

toddkazakov:


Roukoswarf (approved):
looks like it does what it says on the tin.

PR #76, opened 2022-08-24
[BACK-2235] Set newly created clinics `IsMigrated` property to true

clintonium-119:
A number of clinician operations will only work if `clinic.IsMigrated` is true.  This was intended to only apply to empty clinics created for clinician migrations.

For newly clinics created, we set this to true by default so that clinician operations (invite, delete, etc) for the clinic are allowed.


Roukoswarf (approved):
Looks good

PR #78, opened 2022-09-07
[BACK-2249] Only set isMigrated flag for newly created clinics that shouldn't be …

toddkazakov:
…migrated


clintonium-119 (approved):
LGTM :+1:

PR #83, opened 2022-10-14
[BACK-2220] Add Blood Glucose Meter stats

Roukoswarf:
TODO:

- [x] make summary update properly update a summary with only one type filled, not all users have all types, and summaries are updated per-type in platform.


toddkazakov (changes requested):
Filter parsing needs to be improved because the number of filters grew signifficantly

toddkazakov (dismissed):

api/gen_server.go
```diff
@@ -31,9 +31,6 @@ type ServerInterface interface {
 	// Get Clinic by Share Code
 	// (GET /v1/clinics/share_code/{shareCode})
 	GetClinicByShareCode(ctx echo.Context, shareCode string) error
-
-	// (DELETE /v1/clinics/{clinicId})
-	DeleteClinic(ctx echo.Context, clinicId ClinicId) error
```
toddkazakov, line 36: This shouldn't be deleted

toddkazakov (changes requested):

api/gen_server.go
Roukoswarf: Good spot, it was on the todo list, i did not notice when DeleteClinic got merged.

toddkazakov (approved):

toddkazakov (approved):

toddkazakov (approved):

PR #88, opened 2022-12-13
[BACK-2290] Dexcom patient connection

clintonium-119:
[BACK-2290]


patients/service.go
```diff
@@ -69,6 +70,13 @@ func (s *service) Update(ctx context.Context, update PatientUpdate) (*Patient, e
 		}
 	}
 
+	if shouldSetLastRequestedDexcomConnect(*existing, update) {
```
toddkazakov, line 73: Should this be handled in a separate function/endpoint instead of overloading the update? This way it will be similar to how upload reminders are handled.

toddkazakov (approved):

toddkazakov (approved):

PR #99, opened 2023-03-07
[BACK-972] Add the ability to delete empty clinics

toddkazakov:
An empty clinic can contain multiple clinicians, but it doesn't contain any patients or only the demo patient account.


toddkazakov:
Codegen is a bit annoying as it produces non-backward compatible code with minor version updates

Roukoswarf (approved):
LGTM

PR #100, opened 2023-05-04
Add membership restriction endpoints

toddkazakov:


client/client.go
```diff
@@ -1909,6 +1953,87 @@ func NewAssociateClinicianToUserRequestWithBody(server string, clinicId ClinicId
...
+		return nil, err
+	}
+
+	serverURL, err := url.Parse(server)
+	if err != nil {
+		return nil, err
+	}
+
+	operationPath := fmt.Sprintf("/v1/clinics/%s/membership_restrictions", pathParam0)
+	if operationPath[0] == '/' {
```
lostlevels, line 1973: Isn't this always true?

client/client.go
```diff
@@ -1909,6 +1953,87 @@ func NewAssociateClinicianToUserRequestWithBody(server string, clinicId ClinicId
...
+	if err != nil {
+		return nil, err
+	}
+
+	return req, nil
+}
+
+// NewUpdateMembershipRestrictionsRequest calls the generic UpdateMembershipRestrictions builder with application/json body
+func NewUpdateMembershipRestrictionsRequest(server string, clinicId ClinicId, body UpdateMembershipRestrictionsJSONRequestBody) (*http.Request, error) {
+	var bodyReader io.Reader
```
lostlevels, line 1992: stylistic nitpick, ignorable:
If I'm going to read from something I just encoded/wrote, I usually like to use a `bytes.Buffer` / less temp variables.
```go
var buf bytes.Buffer
err := json.NewEncoder(&buf).Encode(body)
... error handling ...
return NewUpdateMembershipRestrictionsRequestWithBody(server, clinicId, "application/json", &buf)
```

client/client.go
```diff
@@ -1909,6 +1953,87 @@ func NewAssociateClinicianToUserRequestWithBody(server string, clinicId ClinicId
...
+		return nil, err
+	}
+
+	serverURL, err := url.Parse(server)
+	if err != nil {
+		return nil, err
+	}
+
+	operationPath := fmt.Sprintf("/v1/clinics/%s/membership_restrictions", pathParam0)
+	if operationPath[0] == '/' {
```
lostlevels, line 2018: Always true?

client/client.go
```diff
@@ -5333,6 +5536,58 @@ func ParseAssociateClinicianToUserResponse(rsp *http.Response) (*AssociateClinic
 	return response, nil
 }
 
+// ParseListMembershipRestrictionsResponse parses an HTTP response from a ListMembershipRestrictionsWithResponse call
+func ParseListMembershipRestrictionsResponse(rsp *http.Response) (*ListMembershipRestrictionsResponse, error) {
+	bodyBytes, err := io.ReadAll(rsp.Body)
+	defer func() { _ = rsp.Body.Close() }()
```
lostlevels, line 5542: Is there a reason to put the `.Close` in a function and not defer it directly?
`defer rsp.Body.Close()`

client/client.go
```diff
@@ -5333,6 +5536,58 @@ func ParseAssociateClinicianToUserResponse(rsp *http.Response) (*AssociateClinic
...
+
+	}
+
+	return response, nil
+}
+
+// ParseUpdateMembershipRestrictionsResponse parses an HTTP response from a UpdateMembershipRestrictionsWithResponse call
+func ParseUpdateMembershipRestrictionsResponse(rsp *http.Response) (*UpdateMembershipRestrictionsResponse, error) {
+	bodyBytes, err := io.ReadAll(rsp.Body)
+	defer func() { _ = rsp.Body.Close() }()
```
lostlevels, line 5568: Same question as other defer call above.

clinics/repo.go
```diff
@@ -340,6 +340,34 @@ func (c *repository) DeletePatientTag(ctx context.Context, id, tagId string) (*C
...
+	clinic, err := c.Get(ctx, id)
+	if err != nil {
+		return nil, err
+	}
+
+	return clinic.MembershipRestrictions, nil
+}
+
+func (c *repository) UpdateMembershipRestrictions(ctx context.Context, id string, restrictions []MembershipRestrictions) error {
+	clinicId, _ := primitive.ObjectIDFromHex(id)
```
lostlevels, line 353: I'm guessing id is already checked before being passed in here so any error can be ignored?

lostlevels (dismissed):
LGTM, had some questions though.

client/client.go
toddkazakov: This is generated

clinics/repo.go
toddkazakov: ids are checked at against the api spec which expects them to be valid ObjectIDs

lostlevels (approved):

PR #104, opened 2023-06-26
[BACK-2551] Bulk tag assign/delete operations

clintonium-119:
[BACK-2551]

To be deployed with:
 - https://github.com/tidepool-org/TidepoolApi/pull/80
 - https://github.com/tidepool-org/clinic-worker/pull/38

[BACK-2551]: https://tidepool.atlassian.net/browse/BACK-2551?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


patients/repo.go
```diff
@@ -95,6 +95,7 @@ func (r *repository) Initialize(ctx context.Context) error {
 		{
 			Keys: bson.D{
 				{Key: "clinicId", Value: 1},
+				{Key: "userId", Value: 1},
```
toddkazakov, line 98: The creation of this index will fail because indexes cannot be updated - you should use a different name.

toddkazakov (changes requested):
Found one small issue with the patient tags index

toddkazakov (approved):

PR #105, opened 2023-06-30
Redox Integration

toddkazakov:


redox/redox.go
```diff
@@ -0,0 +1,399 @@
...
+	if !message.Meta.IsValid() {
+		h.logger.Errorw(
+			"unable to process message with invalid metadata",
+			"metadata", message.Meta,
+		)
+		return fmt.Errorf("%w: cannot process message with invalid metadata", errors.BadRequest)
+	}
+
+	bsonRaw := bson.Raw{}
+	if err := bson.UnmarshalExtJSON(raw, true, &bsonRaw); err != nil {
```
Roukoswarf, line 154: Are we sure this is safe? bson is a superset of json, so it wont crash, but types may not be as expected? String dates, etc? or is there no non-string fields we want out of this?

Roukoswarf: Makes sense to me.

Roukoswarf (approved):

PR #108, opened 2023-08-11
[BACK-2490] Add clinic API which returns TIDE dashboard categories

Roukoswarf:


patients/summary.go
```diff
@@ -0,0 +1,155 @@
...
+	LowGlucoseThreshold      *float64 `bson:"lowGlucoseThreshold,omitempty" json:"lowGlucoseThreshold,omitempty"`
+	SchemaVersion            *int     `bson:"schemaVersion,omitempty" json:"schemaVersion,omitempty"`
+	VeryHighGlucoseThreshold *float64 `bson:"veryHighGlucoseThreshold,omitempty" json:"veryHighGlucoseThreshold,omitempty"`
+	VeryLowGlucoseThreshold  *float64 `bson:"veryLowGlucoseThreshold,omitempty" json:"veryLowGlucoseThreshold,omitempty"`
+}
+
+// PatientSummaryDates dates tracked for summary calculation
+type PatientSummaryDates struct {
+	FirstData         *time.Time `bson:"firstData,omitempty" json:"firstData,omitempty"`
+	HasFirstData      *bool      `bson:"hasFirstData,omitempty" json:"hasFirstData,omitempty"`
```
lostlevels, line 147: More for my understanding but hasFirstData is equivalent to firstData not being empty?

patients/tide.go
```diff
@@ -0,0 +1,56 @@
...
+	TimeCGMUsePercent          *float64     `json:"timeCGMUsePercent,omitempty"`
+	TimeInHighPercent          *float64     `json:"timeInHighPercent,omitempty"`
+	TimeInLowPercent           *float64     `json:"timeInLowPercent,omitempty"`
+	TimeInTargetPercent        *float64     `json:"timeInTargetPercent,omitempty"`
+	TimeInTargetPercentDelta   *float64     `json:"timeInTargetPercentDelta,omitempty"`
+	TimeInVeryHighPercent      *float64     `json:"timeInVeryHighPercent,omitempty"`
+	TimeInVeryLowPercent       *float64     `json:"timeInVeryLowPercent,omitempty"`
+}
+
+type TideResults map[string]*[]TideResultPatient
```
lostlevels, line 54: Is there a reason the value of this map is a pointer to a slice as opposed to a slice?

lostlevels (dismissed):

patients/tide.go
```diff
@@ -0,0 +1,56 @@
...
+
+type PatientTagIds = []string
```
lostlevels, line 56: It his an alias for convenience to not have to cast with the other generated types?

patients/repo.go
```diff
@@ -931,3 +936,233 @@ func cmpToMongoFilter(cmp *string) (string, bool) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		var patientsList []*Patient
+		if err = cursor.All(ctx, &patientsList); err != nil {
+			return nil, fmt.Errorf("error decoding patients list: %w", err)
+		}
+
+		categoryResult := make([]TideResultPatient, 0, 25)
```
lostlevels, line 1071: nitpick: to me, it would seem to make more sense to use a capacity equal to the length of the patientsList rather than an arbritrary number.

patients/repo.go
```diff
@@ -931,3 +936,233 @@ func cmpToMongoFilter(cmp *string) (string, bool) {
...
+				"$gt":  params.CgmLastUploadDateFrom,
+				"$lte": params.CgmLastUploadDateTo,
+			},
+		}
+
+		opts := options.Find()
+		opts.SetLimit(int64(limit))
+
+		opts.SetSort(bson.D{
+			{"summary.cgmStats.periods." + *params.Period + ".timeInTargetPercent", -1},
```
lostlevels, line 1123: I know this is a "catchall" if it hasn't retrieved up to the limit of records yet, but is there a reason why this one sorts -1 but the struct in `categories` with `Field: "timeInTargetPercent"` sorts ascending?

lostlevels (dismissed):
Looks fine, had some questions though.

patients/tide.go
Roukoswarf: yes, and this type has been cut from tide-v2

patients/repo.go
Roukoswarf: that makes sense, done.

patients/tide.go
Roukoswarf: its to match codegen, which generates non-ideal pointer to slice for some reason.

patients/summary.go
Roukoswarf: correct.

patients/repo.go
Roukoswarf: confirmed matches spec

lostlevels (approved):

PR #112, opened 2023-08-24
Make redox verification token optional

toddkazakov:


redox/redox.go
```diff
@@ -28,7 +28,7 @@ const (
 )
 
 type Config struct {
-	VerificationToken string `envconfig:"TIDEPOOL_REDOX_VERIFICATION_TOKEN" required:"true"`
+	VerificationToken string `envconfig:"TIDEPOOL_REDOX_VERIFICATION_TOKEN"`
```
ewollesen, line 31: This seems preferable if it works (I'll verify shortly), since now it will agree with `charts/tidepool/charts/clinic/templates/1-deployment.yaml`, which marks it as optional.

ewollesen (approved):

PR #115, opened 2023-09-13
use make for dependency resolution

ewollesen:
One of make's strengths is its ability to manage dependencies. By creating recipes for each file to be created, installed, or generated, and listing the prerequisites for each of those targets, make can determine if they need to be regenerated, and will handle doing so in the correct order, and where possible, in parallel (if configured).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,11 +15,14 @@ generate:
 	cd client && go generate ./...
 
 # Generate linkerd service profile
-.PHONY: service-profile
-service-profile: generate
-	mkdir -p service-profile
-	openapi-filter -f Confirmations --checkTags spec/clinic.v1.yaml service-profile/clinic.v1.yaml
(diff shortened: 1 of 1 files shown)


ewollesen:
We can take this a lot further, for example, the generate task could be broken down into its individual files. However, it isn't always worth the time spend, and the complexity added to the Makefile. If we find ourselves having to wait for `make generate` a lot, it might be worthwhile.

We could also use some of make's automatic variables, like `$<` (the first prerequisite of the recipe) and `$@` (the target file), but I find those generally more useful in wildcard recipes, which we don't really have here.

ewollesen:
Oh, and by making the recipes the actual files that are created/generated, they don't need to be marked PHONY, and that in turn means that they'll only be run if any one of their prerequisites have a newer file modification time.

toddkazakov:
Thank you!

PR #117, opened 2023-09-18
TIDE Pagination

Roukoswarf:

PR #127, opened 2023-11-08
[BACK-2744] Add AnyHigh and AnyLow to Summaries

Roukoswarf:


ewollesen (approved):

patients/tide.go
```diff
@@ -25,7 +25,7 @@ type TideConfig struct {
 type TideFilters struct {
 	DropInTimeInTargetPercent *string `json:"dropInTimeInTargetPercent,omitempty"`
 	TimeCGMUsePercent         *string `json:"timeCGMUsePercent,omitempty"`
-	TimeInLowPercent          *string `json:"timeInLowPercent,omitempty"`
+	TimeInAnyLowPercent       *string `json:"timeInAnyLowPercent,omitempty"`
```
ewollesen, line 28: This again could just be my ignorance, but should there be a `TimeInAnyHighPercent`?

client/types.go
```diff
@@ -1128,6 +1162,42 @@ type PatientCGMPeriod struct {
...
+	// TimeInAnyHighRecordsDelta Difference between the timeInAnyHighRecords in this period and version in the opposite offset
+	TimeInAnyHighRecordsDelta *int `json:"timeInAnyHighRecordsDelta,omitempty"`
+
+	// TimeInAnyLowMinutes Counter of minutes spent in Any low glucose range
+	TimeInAnyLowMinutes *int `json:"timeInAnyLowMinutes,omitempty"`
+
+	// TimeInAnyLowMinutesDelta Difference between the timeInAnyLowMinutes in this period and version in the opposite offset
+	TimeInAnyLowMinutesDelta *int `json:"timeInAnyLowMinutesDelta,omitempty"`
+
+	// TimeInAnyLowPercent Percentage of time spent in Any low glucose range
```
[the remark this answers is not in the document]
ewollesen: Ah, from https://github.com/tidepool-org/clinic/pull/127/commits/208585ae07e8f1df7377df8cbfbc3151bee16b78#diff-41f8b0b50297a7227a61ed2ada54e8c8d28b761983d650b2b1b53fcca1b8a76cR335 I see that it's 0-1.

client/client.go
ewollesen, line 1: I'm a fan of generated files being named foo.gen.go or foo-gen.go or foo_gen.go or... you get the idea. I find it helps to know just at a glance that a file is generated. It helps when reviewing, or dealing with merge conflicts, wondering why there are no tests, etc.

Not a big deal, just something to think about.

Oh, and I'll add that while there's a header in the file, you have to beat GitHub with a stick to get it to actually show you that header, so it's not always obvious from the file header that a file has been generated. How the heck isn't there a "show me the whole damn file" button?

client/types.go
```diff
@@ -980,6 +984,30 @@ type PatientBGMPeriod struct {
 	HasTimeInVeryLowRecords  *bool    `json:"hasTimeInVeryLowRecords,omitempty"`
 	HasTotalRecords          *bool    `json:"hasTotalRecords,omitempty"`
 
+	// TimeInAnyHighPercent Percentage of time spent in Any high glucose range
```
ewollesen, line 987:
```suggestion
	// TimeInAnyHighPercent Percentage of time spent in any high glucose range
```

ewollesen (commented):
Mostly just questions, assuming those questions don't turn up issues, then I think this is GTG.

api/mappers.go
```diff
@@ -1173,9 +1235,11 @@ func ParseCGMSummaryFilters(params ListPatientsParams) (filters patients.Summary
 		"timeInTargetPercent":   params.CgmTimeInTargetPercent,
 		"timeInHighPercent":     params.CgmTimeInHighPercent,
 		"timeInVeryHighPercent": params.CgmTimeInVeryHighPercent,
+		"timeInAnyHighPercent":  params.CgmTimeInAnyHighPercent,
```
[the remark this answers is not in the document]
Roukoswarf: nope, looks like a typo, good catch.

api/mappers.go
```diff
@@ -1173,9 +1235,11 @@ func ParseCGMSummaryFilters(params ListPatientsParams) (filters patients.Summary
...
 
 		"timeCGMUseRecords":     params.CgmTimeCGMUseRecords,
 		"timeInVeryLowRecords":  params.CgmTimeInVeryLowRecords,
+		"timeInAnyLowRecords":   params.CgmTimeInAnyLowRecords,
```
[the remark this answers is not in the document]
Roukoswarf: good catch, missed it.

patients/tide.go
Roukoswarf: This is just the filters struct we send back via the API for the client to know the filters, its an early version of what it should be and just hardcodes in the values that were used for the categories.

We dont have a category using the `AnyHighPercent`, so it doesn't exist here for the API response.

A more flexible version already exists in the #117 PR.

client/client.go
Roukoswarf: We do control the generation output file name, as long as theyre in the same package they will work, so i could get behind a naming change for clarity.

Theyre generated here: https://github.com/tidepool-org/clinic/blob/master/Makefile#L5-L15

ewollesen (dismissed):
LGTM

I'm sure you're aware, but just in case, make sure to have QA do a full regression, in case something got missed with all the pointer changes. :)

PR #130, opened 2023-11-10
[BACK-2703] increase max patient tags to 50

ewollesen:


clinics/repo.go
```diff
@@ -462,16 +462,16 @@ func (c *repository) UpdateMRNSettings(ctx context.Context, id string, settings
 	return err
 }
 
-func canAddPatientTag(clinic Clinic, tag PatientTag) (bool, error) {
+func canAddPatientTag(clinic Clinic, tag PatientTag) error {
```
toddkazakov, line 465: nit: `can___` implies the function returns a boolean. Consider prefixing this with `assert`

toddkazakov (dismissed):

clinics/repo.go
ewollesen: Updated

toddkazakov (approved):

PR #136, opened 2024-01-15
[BACK-2693] Enforce patient count limit

darinkrauss:
- Implement patient count APIs
- Implement patient count settings APIs
- Maintain patient count after patient create and remove operations
- Enforce patient count limit before patient create
- Use background context in tests rather than nil
- Add and update tests
- Update air configuration
- Update gitignore and dockerignore
- Update Makefile
- https://tidepool.atlassian.net/browse/BACK-2693


patients/service.go
```diff
@@ -32,31 +34,45 @@ func (s *service) Get(ctx context.Context, clinicId string, userId string) (*Pat
...
+}
+
 func (s *service) List(ctx context.Context, filter *Filter, pagination store.Pagination, sorts []*store.Sort) (*ListResult, error) {
 	return s.repo.List(ctx, filter, pagination, sorts)
 }
 
-func (s *service) Create(ctx context.Context, patient Patient) (*Patient, error) {
-	if err := s.enforceMrnSettings(ctx, patient.ClinicId.Hex(), patient.UserId, &patient); err != nil {
-		return nil, err
+func (s *service) Create(ctx context.Context, patient Patient) (*Patient, bool, error) {
```
toddkazakov, line 45: There's no way to tell what does the boolean argument mean without actually reading the implementation of the function. When returning more than 2 arguments, it's best to group those in a struct and return that struct. This way the code will be easier to read because the boolean will now be returned in a named field. 

In this specific example, I'm not entirely sure it's necessary to return that boolean argument. If there's no error - the patient was created successfully. I'll leave it up to you to decide whether that's really needed or not, but if it is, it ought to be returned in a result struct so we can have only 2 return arguments.

patients/service.go
```diff
@@ -240,6 +269,61 @@ func (s *service) enforceMrnSettings(ctx context.Context, clinicId string, exist
...
+	if patientCountSettings.HardLimit == nil {
+		return nil
+	}
+
+	// If outside start date and end date, if specified, then allow
+	now := time.Now()
+	if patientCountSettings.HardLimit.StartDate != nil && now.Before(*patientCountSettings.HardLimit.StartDate) {
+		return nil
+	} else if patientCountSettings.HardLimit.EndDate != nil && now.After(*patientCountSettings.HardLimit.EndDate) {
+		return nil
```
toddkazakov, line 295: Shouldn't we prevent adding patients after the current plan expires, instead of allowing it?

toddkazakov (changes requested):
Nothing major - left inline comments

patients/service.go
darinkrauss: The reason this function returns the extra boolean is because the underlying Repository needs to and it shares the same interface as the Service. The Repository needs to return it for two reasons:
- The Repository may or may not actually create a new Patient in the database. The Repository did not distinguish between creating a user and simply updating a user that already existed. It just returned the Patient. With the addition of the boolean, it is now returns whether the Patient was created or not so that...
- The Service needs to know if a Patient was created or not so that it can recalculate the patient count and set that on the Clinic, if necessary.

I could remove the boolean and just blindly recalculate the patient count regardless of whether a new Patient was actually added to the database or not. What do you think?

Also, this applies to the two `Delete*` operations. That is, the addition of the boolean return value from the Repository so that the Service can know whether to recalculate the patient count or not.

Oh, and as to why the Patient Repository cannot do the recalculation, the Repository does not have the ability to inform the Clinic service of the new patient count. Only the Patient Service has that ability. This is due to the design of the interfaces and the dependency ordering.

toddkazakov: > The Repository may or may not actually create a new Patient in the database. The Repository did not distinguish between creating a user and simply updating a user that already existed. It just returned the Patient.

This logic seems to be misplaced, it should have been in the service. No need to fix this now.

> I could remove the boolean and just blindly recalculate the patient count regardless of whether a new Patient was actually added to the database or not. What do you think?

A patient will be updated in `repo.Create` rarely - only when merging legacy clinician account with a clinic. It's pretty safe to assume that there won't be any negative performance impact. Blindly recalculating seems simpler.

patients/service.go
darinkrauss: The hard limit only applies if the current time (`now`) is *within* the start and end dates, if specified. So, in this case, if the `EndDate` is specified and `now` is after the `EndDate`, then the limit is NOT applied, so returns `nil` here.

patients/service.go
darinkrauss: @toddkazakov I can make that change. Should I make the change for the two `Delete*` operations as well. That is, blindly recalculate the patient count?

patients/service.go
toddkazakov: Right, that was my understanding as well. However, I don't understand why would we want to have an end date for the hard limit.

From the ticket description it seems that we should capture a start and end dates for the current subscription of a clinic. We can't use the hard limit start and end dates for this, because we would be relaxing the restrictions when the plan expires.

patients/service.go
toddkazakov: Feel free to handle the delete operations as you see fit - my initial comment was regarding the signature of the function. I've always considered having more than 2 return arguments a code smell.

darinkrauss: The `Create` function has been updated to remove the extra boolean return value and unconditionally recalculate the patient count per above comments.

patients/service.go
toddkazakov: Just wanted to make sure the end date won't be used as the plan end date, because I don't think we should relax that patient count limit when the plan expires.

toddkazakov (dismissed):

patients/service.go
darinkrauss: Yes, I agree. 👍

darinkrauss:
Replaced by https://github.com/tidepool-org/clinic/pull/145.

PR #137, opened 2024-01-17
BACK-2813 Accept full name, birth date, mrn and tags when creating a …

toddkazakov:
…patient


patients/users.go
```diff
@@ -95,7 +95,7 @@ func (s *userService) GetUser(userId string) (*shoreline.UserData, error) {
 	return user, nil
 }
 
-func (s *userService) GetPatientFromExistingUser(ctx context.Context, patient *Patient) error {
+func (s *userService) PopulatePatientDetailsFromExistingUser(ctx context.Context, patient *Patient) error {
```
ewollesen, line 98: I like this rename.

ewollesen (approved):

PR #144, opened 2024-04-15
BACK-2957 PROD-85 Evaluate a clinic merge and generate report

toddkazakov:


clinics/merge/cluster.go
```diff
@@ -0,0 +1,235 @@
...
+			patient, err := p.graph.Vertex(id)
+			if err != nil {
+				return nil, err
+			}
+
+			conflicts := map[string][]string{}
+			for duplicate, edge := range adjecencyMap[id] {
+				q.Add(duplicate)
+				conflict := edge.Properties.Attributes[conflictAttributeKey]
+				conflicts[conflict] = append(conflicts[conflict], duplicate)
```
lostlevels, line 107: For my understanding, if user `"A`" had a `"MRN Only Match`" conflict with user `"B`", and `"B`" has a `"Name Only Match`" conflict with user `"C`", `"A`", `"B`", and `"C`" would be part of the same cluster right?

toddkazakov: Correct

clinics/merge/patients.go
```diff
@@ -0,0 +1,221 @@
...
+		source:         source,
+		sourcePatients: sourcePatients,
+		sourceTags:     buildTagsMap(source.PatientTags),
+		target:         target,
+		targetByUserId: make(map[string]*patients.Patient),
+		targetPatients: targetPatients,
+		targetTags:     buildTagsMap(target.PatientTags),
+	}
+	for _, patient := range targetPatients {
+		planner.targetByUserId[getUserId(patient)] = &patient
```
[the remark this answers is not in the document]
toddkazakov: Experimental support was added in 1.21 and it's GA in 1.22

clinics/merge/plan.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"context"
+)
+
+type Plan interface {
+	PreventsMerge() bool
+}
+
+type Planner[T Plan] interface {
```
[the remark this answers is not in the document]
toddkazakov: The only reason is to be able to run multiple planners and return a typed list of plans. This code would have been duplicated if the planner wasn't generic:
```
func RunPlanners[T Plan](ctx context.Context, planners []Planner[T]) ([]T, error) {
	result := make([]T, 0, len(planners))
	for _, planner := range planners {
		res, err := planner.Plan(ctx)
		if err != nil {
			return nil, err
		}
		result = append(result, res)
	}
	return result, nil
}
```

clinics/merge/tags.go
```diff
@@ -0,0 +1,123 @@
...
+
+func NewSourceTagMergePlanner(tag clinics.PatientTag, source, target clinics.Clinic) Planner[TagPlan] {
+	return &SourceTagMergePlanner{
+		tag:    tag,
+		source: source,
+		target: target,
+	}
+}
+
+func (t *SourceTagMergePlanner) CanRun() bool {
```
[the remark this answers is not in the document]
toddkazakov: No, I removed it

clinics/merge/plan.go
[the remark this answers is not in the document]
lostlevels: Since the actual types do different things I would prefer a function over an interface and explicitly stating the return type for ease of reading. Something along the lines of:

```go
func runPlanners[T any](planners ...func(ctx context.Context) (T, error)) ([]T, error) {
	var res []T
	for _, runner := range planners {
		r, err := runner(context.Background())
		if err != nil {
			return nil, err
		}
		res = append(res, r)
	}
	return res, nil
}


type intermediatePlanner struct {
	SettingsPlanners []func(context.Context) (SettingsPlan, error)
	TagPlanners      []func(context.Context) (TagPlan, error)
	...
}

func (m *ClinicMergePlanner) Plan(ctx context.Context) (plan ClinicMergePlan, err error) {
	intermediate.SettingsPlanners = []func(context.Context) (SettingsPlan, error) {
		NewSettingsReporterPlanner(source, target, GetMRNRequiredSettings, TaskTypeClinicSettingsMRNRequired).Plan,
		NewSettingsReporterPlanner(source, target, GetMRNUniqueSettings, TaskTypeClinicSettingsMRNUnique).Plan,
		...
	}
}
```

but that might not be worth the effort at this point.

clinics/merge/plan.go
```diff
@@ -0,0 +1,34 @@
...
+		res, err := planner.Plan(ctx)
+		if err != nil {
+			return nil, err
+		}
+		result = append(result, res)
+	}
+	return result, nil
+}
+
+func PlansPreventMerge[T Plan](plans []T) bool {
```
lostlevels, line 27: Please avoid unneeded type parameters / generic functions as it makes the code harder to use. Use the interface `Plan` type directly.

```
func PlansPreventMerge(plans []Plan) bool {
```

lostlevels (changes requested):

clinics/merge/plan.go
toddkazakov: Unfortunately that's not possible without iterating over the arrays with concrete implementations, and creating a new array with where each structs is cast to `Plan`. For example, `[]ClinicianPlan` cannot be be passed to a function which accepts a `[]Plan` parameter.

lostlevels: I see what you're saying.

lostlevels (dismissed):

PR #146, opened 2024-04-23
[BACK-3010] Allow adding service accounts to clinic workspaces

toddkazakov:
Adds an endpoint which creates a clinician with a role `CLINIC_MEMBER` given `client_id` and `client_secret` of a client which supports the client credentials grant. This way, partner organizations can programatically access clinic workspaces easier. The current approach was to ask clinic admins to manually register and add a clinician account and obtain access/refresh tokens using the password credentials which is not suitable for this flow.

The client credentials grant is more appropriate, because it allows acting on behalf of the client. In Keycloak, this is achieved by using service accounts - user account without an email/password. 

The clinic service authenticates with the credentials provided in the request to obtain the user id of the service account from the returned `id_token`.


auth/serviceaccount.go
```diff
@@ -0,0 +1,60 @@
...
+
+type ServiceAccount struct {
+	UserId string
+}
+
+type ServiceAccountAuthenticator struct {
+	TokenEndpoint string `envconfig:"TIDEPOOL_AUTH_SERVICE_TOKEN_ENDPOINT"`
+}
+
+func NewServiceAccountAuthenticator() (*ServiceAccountAuthenticator, error) {
```
ewollesen, line 19: Have you considered pulling envconfig out of this function? It would be easier to test, and for other things to use, if the arguments it needed were passed to it, as opposed to needing to be mocked via envconfig.

As a reader of the code, I'd prefer not to have to know anything about envconfig in order to know how the properties of ServiceAccountAuthenticator are being established.

client/mock.go
ewollesen, line 1: I know that files ending in _test.go aren't included in final binaries... Would this file be removed due to tree-shaking or something?

Not a big deal, and I'm mostly just curious, but if we don't need to ship files like this, then they can't come back to haunt us, and if it makes the binary smaller, even better.

Maybe an interesting experiment to compare the resulting binary sizes with this as is, versus renaming this file to mock_test.go or something.

clinicians/clinicians.go
```diff
@@ -14,7 +14,8 @@ var (
 	ErrNotFound  = fmt.Errorf("clinician %w", errors.NotFound)
 	ErrDuplicate = fmt.Errorf("%w: clinician is already a member of the clinic", errors.Duplicate)
 
-	ClinicAdmin = "CLINIC_ADMIN"
+	ClinicAdmin  = "CLINIC_ADMIN"
```
ewollesen, line 17: Not a big deal, but I'm curious as to why these aren't constants?

integration/integration_suite_test.go
```diff
@@ -1,16 +1,90 @@
...
+func setupEnvironment() {
+	dbTest.SetupDatabase()
+
+	authStub = integrationTest.AuthStub()
+	seagullStub = integrationTest.SeagullStub()
+	shorelineStub = integrationTest.ShorelineStub()
+	xealthStub = xealthTest.ServerStub()
+	keycloakStub := integrationTest.KeycloakStub()
+
+	Expect(os.Setenv("LOG_LEVEL", "error")).To(Succeed())
```
ewollesen, line 45: I suspect you'd be better off using [testing.T.Setenv](https://cs.opensource.google/go/go/+/refs/tags/go1.21.1:src/testing/testing.go;l=1421) here..?

I don't know if ginkgo supports it via GinkgoT() or not though..?

But it's useful because it handles cleanup of the variables after the test for you.

integration/integration_suite_test.go
```diff
@@ -1,16 +1,90 @@
...
+	go func() {
+		_ = app.Start(context.Background())
+	}()
+
+	wg.Wait()
+}
+
+func teardownEnvironment() {
+	dbTest.TeardownDatabase()
+	os.Clearenv()
```
ewollesen, line 83: This is a footgun that could easily come back to bite you, and could affect other tests (that's kinda an indication that other tests are poorly written, but still). Ask me about the time I found out a system I was working on had a caching layer for env variable lookups (no really!)

I don't think it's terribly strange that a test should be able to expect basic env vars to exist, like $HOME and $USER.

integration/test/stubs.go
```diff
@@ -124,3 +133,52 @@ func AuthStub() *httptest.Server {
...
 	}))
 }
+
+func KeycloakStub() *httptest.Server {
+	return httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		var resp []byte
+		if r.Method == http.MethodPost && r.RequestURI == "/realms/integration-test/protocol/openid-connect/token" {
+			if err := r.ParseForm(); err != nil {
+				w.WriteHeader(http.StatusNotImplemented)
+			} else {
```
ewollesen, line 143: if you return after the status not implemented, you can dedent the rest of this, and make the logic a little easier to read.

spec/clinic.v1.yaml
```diff
@@ -1812,6 +1812,25 @@ paths:
...
       x-internal: true
     parameters: []
+  '/v1/clinics/{clinicId}/service_accounts':
+    parameters:
+      - $ref: '#/components/parameters/clinicId'
+    post:
+      summary: Add Service Account
+      operationId: AddServiceAccount
+      responses:
+        '200':
```
ewollesen, line 1822: I don't think 200 is the only possible response..?

...or is this a generated file? I wonder because of all the stoplight ids.

spec/clinic.v1.yaml
```diff
@@ -4164,6 +4183,31 @@ components:
           minLength: 1
         tags:
           $ref: '#/components/schemas/PatientTagIds'
+    AddServiceAccount:
```
ewollesen, line 4186: I don't know how much we care, but is required a better flag here than minLength 1? I guess it doesn't matter as we're not specifying any details at all really anyway.

store/test/database.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 const (
-	mongoTimeout = time.Second * 5
+	mongoTimeout = time.Second * 15
```
ewollesen, line 18: Were you seeing timeouts? Do we have any metrics on that (i.e. how often and what calls were timing out?)

ewollesen (dismissed):
Nothing blocking, just a few comments.

The stubs (for keycloak and other services) feel like they're going to get out of hand with spaghetti logic. 

Certainly we want to keep them simple and flexible, but there's got to be a better way we can come up with?

*Update*: Yeah, that's right, go 1.22 added enhanced pattern matching:

https://pkg.go.dev/net/http#hdr-Patterns

Which could probably be used to good effect in the stubs.

auth/serviceaccount.go
toddkazakov: The purpose of this function is to instantiate the struct with the configuration from the env. If you want to instantiate it without knowing anything about envconfig you can always do:
```golang
authenticator := &ServiceAccountAuthenticator{
  TokenEndpoint: "https://...."
}
```

clinicians/clinicians.go
toddkazakov: Good catch

spec/clinic.v1.yaml
toddkazakov: It is required at the moment and must not be empty.

store/test/database.go
toddkazakov: The teardown was timing out. I've seen it happen few times locally, but never in CI.

spec/clinic.v1.yaml
toddkazakov: This is not a generated file, but it was edited with the Stoplight editor. 200 is the only valid response and given this is for internal use only (will be called from ORCA or manually using CURL) I think it's good enough. Does documenting 401 or 500 provide any benefit?

integration/integration_suite_test.go
toddkazakov: TIL this exists, thanks!

integration/integration_suite_test.go
toddkazakov: `GinkgoT().SetEnv` handles the cleanup now, so this code is removed. Thanks!

integration/test/stubs.go
toddkazakov: Is the call to `w.Write()` not needed? I believe it is.

toddkazakov:
>The stubs (for keycloak and other services) feel like they're going to get out of hand with spaghetti logic.

Valid critique, but probably out of scope. There's may be an elegant way to use the builder pattern to simplify how stubbed requests are handled but there is some fine balance between adding fast and easy stubbed out responses vs reimplementing the dependency in-memory. 

@ewollesen Ready for final review

auth/serviceaccount.go
ewollesen: True. I still think it's a strange constructor, and not at all what I would expect. I'd expect a `New` to do a basic construction without envconfig, and a `NewFromEnvconfig` to do what you've done here. But since there's no other real initialization of maps or channels going on, this is fine.

integration/integration_suite_test.go
ewollesen: > TIL this exists, thanks!

It's not super old (go 1.17).

spec/clinic.v1.yaml
ewollesen: > Does documenting 401 or 500 provide any benefit?
Not really, but it's done in most of the API endpoints' documentation, so I thought it was something the team did.

spec/clinic.v1.yaml
ewollesen: Are you telling me that something that's required can be empty? Seems weird to me. But yeah, I just double checked the docs, and they make no mention of it not being possible to send an empty string when a property is marked required. Seems weird, but there you go.

auth/serviceaccount.go
```diff
@@ -0,0 +1,60 @@
...
+	s := &ServiceAccountAuthenticator{}
+	err := envconfig.Process("", s)
+	return s, err
+}
+
+func (s *ServiceAccountAuthenticator) GetServiceAccount(ctx context.Context, clientId, clientSecret string) (*ServiceAccount, error) {
```
ewollesen, line 25: This function has no test coverage.

Given what it does, I don't think a test is necessary. Maybe the next time you're changing this code, consider throwing this a test that checks that the returned ServiceAccount's UserId isn't blank (that's really about the only logic in here).

The test coverage of authentication.go is also 0, and that seems bad, but out of scope for this PR.

integration/test/stubs.go
ewollesen: It's only needed if there's a response body, and in this case, I don't believe there is.

integration/xealth_test.go
```diff
@@ -667,6 +604,10 @@ func asServer(req *http.Request) {
 	req.Header.Set("x-tidepool-session-token", integrationTest.TestServerToken)
 }
 
+func asServiceAccount(req *http.Request) {
+	req.Header.Set("x-tidepool-session-token", integrationTest.TestServiceAccountToken)
```
ewollesen, line 608: x-tidepool-session-token exists as both a var and a const in the auth package:

auth/authentication.go:18
auth/authorizer.go:23

It'd probably be best to settle on one usage (an exported const would be my preference).

That said, this code is accurate and working, so not a blocker.

ewollesen (approved):
No blockers, just food for thought.

integration/xealth_test.go
toddkazakov: An export const won't catch an accidental change of the header key. If we had an exported const in the auth package e.g. `TidepoolSessionTokenKey` and we had both the implementation and tests use the value, we could set the value to an arbitrary string (e.g. `x-session-token`) and tests would still pass. However, this will be a breaking change because existing clients use `x-tidepool-session-token`.

PR #148, opened 2024-06-03
[BACK-2677] EHR PDF on Upload

toddkazakov:
- Adds a new endpoint to reschedule a single patient order
- Adds EHR configuration to enable to enable EHR PDF reports on upload and which model to use
- Adds integration tests for Redox subscriptions

This PR is stacked on top of #146 to leverage refactored integration test suite


Roukoswarf (approved):
Nothing concerning that I can see.

PR #149, opened 2024-06-12
[BACK-2963] Standard Deviation and other new CGM stats

Roukoswarf:


ewollesen (dismissed):
LGTM

ewollesen (commented):
I don't have any issues with this, but I'm curious why the go version is being bumped to 1.22.4? There's no explanation in the commit message. All of our other projects are using 1.22.

I don't think it hurts anything to bump it... I'd probably use 1.22.x if it were up to only me, but since we've decided to hard code, it seems like a bad idea to have a bunch of different versions around? I dunno... Especially given that 1.22.5 is current as of time of writing.

ewollesen (dismissed):

ewollesen (approved):

PR #154, opened 2024-07-30
[BACK-2958] [BACK-2960] [BACK-3076] Add ability to execute clinic merge plans

toddkazakov:


clinics/manager/manager.go
```diff
@@ -219,6 +224,32 @@ func (c *manager) getDemoPatient(ctx context.Context) (*patients.Patient, error)
...
+func (c *manager) deleteClinic(ctx context.Context, clinicId string) error {
+	filter := patients.Filter{ClinicId: &clinicId}
+	pagination := store.Pagination{Limit: 2}
+	res, err := c.patientsService.List(ctx, &filter, pagination, nil)
+
+	if err != nil {
+		return err
+	}
+	if res == nil {
+		return fmt.Errorf("patient list result not defined")
```
lostlevels, line 236: Is there a reason sometimes the error is defined in the `errors` package and sometimes defined inline?

clinics/manager/manager.go
```diff
@@ -219,6 +224,32 @@ func (c *manager) getDemoPatient(ctx context.Context) (*patients.Patient, error)
...
+	}
+	if !c.patientListAllowsClinicDeletion(res.Patients) {
+		return fmt.Errorf("%w: deletion of non-empty clinics is not allowed", errors.BadRequest)
```
lostlevels, line 239: Same as above, Is there a reason this "constant" error is wrapped inline and not defined as a variable in the `errors` package?

lostlevels (dismissed):

clinics/manager/manager.go
toddkazakov: The errors in the `errors` package ought to wrapped so when they bubble up to the API layer we return the correct http code. All other errors which bubble up with result in a 500 response, which is intended here.

lostlevels (approved):

PR #161, opened 2024-10-08
[BACK-3191] Keep track of patient deletions

toddkazakov:
The patient service tests are hideous and are testing specific implementation details because of the mocks and should be rewritten to use the repositories directly (and make them integration tests by definition), but this is out of the scope of this PR.


patients/service_test.go
```diff
@@ -3,6 +3,7 @@ package patients_test
 import (
 	"context"
 	"fmt"
+	test2 "github.com/tidepool-org/clinic/store/test"
```
darinkrauss, line 6: `clinicStoreTest`? Also, perhaps order imports?

patients/service_test.go
```diff
@@ -27,18 +28,24 @@ var _ = Describe("Patients Service", func() {
 	var service patients.Service
 	var clinicsService *clinicsTest.MockService
 	var repo *patientsTest.MockRepository
+	var deletionsRepo *patientsTest.MockDeletedPatientsRepository
+	var deletedRepoCtrl *gomock.Controller
```
darinkrauss, line 32: `deleted` -> `deletions`?

patients/service.go
```diff
@@ -152,7 +188,7 @@ func (s *service) DeletePermission(ctx context.Context, clinicId, userId, permis
 			"deleting patient from clinic because the patient revoked all permissions",
 			"userId", userId, "clinicId", clinicId,
 		)
-		if err := s.Remove(ctx, clinicId, userId); err != nil {
+		if err := s.Remove(ctx, clinicId, userId, &userId); err != nil {
```
darinkrauss, line 191: Is the `userId` the id of the user that actually performed the deletion?

patients/service.go
```diff
@@ -137,13 +173,13 @@ func (s *service) UpdatePermissions(ctx context.Context, clinicId, userId string
 			"deleting patient from clinic because the patient revoked all permissions",
 			"userId", userId, "clinicId", clinicId,
 		)
-		return nil, s.Remove(ctx, clinicId, userId)
+		return nil, s.Remove(ctx, clinicId, userId, &userId)
```
darinkrauss, line 176: Is the userId the id of the user that actually performed the deletion?

patients/service.go
```diff
@@ -103,26 +109,56 @@ func (s *service) Update(ctx context.Context, update PatientUpdate) (*Patient, e
...
 }
 
 func (s *service) UpdateEmail(ctx context.Context, userId string, email *string) error {
 	s.logger.Infow("updating patient email", "userId", userId, "email", email)
-	return s.repo.UpdateEmail(ctx, userId, email)
+	return s.patientsRepo.UpdateEmail(ctx, userId, email)
 }
 
-func (s *service) Remove(ctx context.Context, clinicId string, userId string) error {
+func (s *service) Remove(ctx context.Context, clinicId string, userId string, clinicianId *string) error {
```
darinkrauss, line 127: `clinicianId` -> `deletedByUserId`?

patients/deletions_repo.go
```diff
@@ -0,0 +1,74 @@
...
+			},
+			Options: options.Index().
+				SetName("DeletedTime"),
+		},
+	})
+	return err
+}
+
+func (p *patientDeletionsRepository) Create(ctx context.Context, deletion PatientDeletion) error {
+	if deletion.DeletedTime.IsZero() {
```
darinkrauss, line 67: Do we want to check that the patient id is not nil?

darinkrauss (changes requested):
A bunch of minor naming issues, but one or two necessary changes.

patients/service.go
toddkazakov: In this case we delete the patient because they revoked the permissions, so it makes sense to log that user id

patients/deletions_repo.go
toddkazakov: Being more defensive won't hurt

patients/test/mock_deleted_repository.go
```diff
@@ -0,0 +1,50 @@
+// Code generated by MockGen. DO NOT EDIT.
```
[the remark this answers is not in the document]
darinkrauss: Looking at the latest code, this file is generated as `mock_deletions_repository.go`. Suggest you delete this file (named `mock_deleted_repository.go` - note `deleted` vs. `deletions`) and run generate again.

patients/test/mock_repository.go
```diff
@@ -200,17 +200,17 @@ func (mr *MockRepositoryMockRecorder) List(ctx, filter, pagination, sort interfa
 }
 
 // Remove mocks base method.
-func (m *MockRepository) Remove(ctx context.Context, clinicId, userId string) error {
+func (m *MockRepository) Remove(ctx context.Context, clinicId, userId string, deletedUserId *string) error {
```
darinkrauss, line 203: [nit] `deletedByUserId`?

patients/test/mock_service.go
```diff
@@ -200,17 +200,17 @@ func (mr *MockServiceMockRecorder) List(ctx, filter, pagination, sort interface{
 }
 
 // Remove mocks base method.
-func (m *MockService) Remove(ctx context.Context, clinicId, userId string) error {
+func (m *MockService) Remove(ctx context.Context, clinicId, userId string, deletedUserId *string) error {
```
darinkrauss, line 203: [nit] `deletedByUserId`?

patients/patients.go
```diff
@@ -94,6 +94,13 @@ func (p Patient) IsCustodial() bool {
 	return p.Permissions != nil && p.Permissions.Custodian != nil
 }
 
+type PatientDeletion struct {
```
darinkrauss, line 97: [nit] Its up to you what you want to name this, but, as commented before, consider just `Deletion` since it is in the `patients` package. I see you made some changes, but didn't get the others. Perhaps a case-insensitive search on "patientdeletions" and review for other changes? Entirely up to you, though.

darinkrauss (changes requested):
A couple of minor naming suggestions (but up to you). I think fixing the generated file is a good idea though. Let me know if I'm missing something, though.

toddkazakov:
I forgot to regenerate the mock, apologies. I believe I fixed all of the naming issue you pointed out.

darinkrauss (approved):
LGTM!

PR #167, opened 2024-12-04
[BACK-3254] Prevent merge if there are duplicate MRNs

toddkazakov:


clinics/merge/errors.go
```diff
@@ -0,0 +1,23 @@
+package merge
+
+import (
+	"errors"
+	mapset "github.com/deckarep/golang-set/v2"
+)
+
+type Error error
```
lostlevels, line 8: Is there a reason to make this a distinct type?

lostlevels (dismissed):

clinics/merge/errors.go
toddkazakov: The idea was to prevent accidentally returning internal go errors in reports, but I see your point that this doesn't work because it references the `error` interface. I have changed this to be a regular struct.

lostlevels (approved):

lostlevels (approved):

PR #172, opened 2025-01-24
[BACK-3765] Handle summary update being reverted by patient update, allow 0 character summary ID

Roukoswarf:
Summaries updated via Update Patient would cause unintended overwrites of summary data


ewollesen (dismissed):
LGTM

toddkazakov (changes requested):
This ought to be handled in the API layer. Please also update the docs and set the summary attribute as read only.

toddkazakov (dismissed):

PR #174, opened 2025-02-13
BACK-2538 Retain deleted clinic and clinician records

toddkazakov:


deletions/deletions_repo.go
```diff
@@ -0,0 +1,117 @@
...
+
+	if _, err := p.collection.InsertMany(ctx, documents); err != nil {
+		return fmt.Errorf("error persisting deleted objects in collection %s: %w", p.collection.Name(), err)
+	}
+	return nil
+}
+
+func (p *deletionsRepository[T]) prepareDocument(deleted T, meta Metadata) bson.M {
+	deletion := bson.M{
+		"deletedTime":   time.Now(),
```
lostlevels, line 110: I almost feel logically the `Metadata` type should include the deleted time. Though I suppose that'd be tedious to include with every creation of Metadata.

lostlevels (dismissed):

PR #179, opened 2025-04-08
[BACK-3517] Add the total count of patients in the clinic to patient list responses

toddkazakov:


Makefile
```diff
@@ -14,9 +14,11 @@ NPM_PKG_SPECS = \
 
 PATH:=$(shell pwd)/$(TOOLS_BIN):$(PATH)
 
+OSTYPE:=$(shell uname -msr)
```
Roukoswarf, line 17: `OSTYPE` is an unexported shell builtin, and this is my mistake. Using uname for this might not be stable output.

`OSTYPE:=$(shell echo $${OSTYPE})` is a bit roundabout, but works for me and might be worth a test.

This should output lowercase `darwin*` on macos.

Roukoswarf (changes requested):
Code itself looks good, one suggestion to try RE macos makefile.

Roukoswarf (dismissed):

api/patients.go
```diff
@@ -60,7 +60,14 @@ func (h *Handler) ListPatients(ec echo.Context, clinicId ClinicId, params ListPa
 		return err
 	}
 
-	return ec.JSON(http.StatusOK, NewPatientsResponseDto(list))
+	clinicPatientsCount, err := h.Patients.Count(ctx, &patients.Filter{
```
tjotala, line 63: Would it be easier/cheaper to go fetch the count available in the corresponding clinic object? https://github.com/tidepool-org/clinic/blob/master/clinics/clinics.go#L105

toddkazakov: It will be cheaper, but incorrect. The count in the clinic object excludes the demo patient account (James Jellyfish).

api/mappers.go
```diff
@@ -579,11 +579,14 @@ func NewPatientsDto(patients []*patients.Patient) []Patient {
...
 
-func NewPatientsResponseDto(list *patients.ListResult) PatientsResponse {
-	data := Patients(NewPatientsDto(list.Patients))
+func NewPatientsResponseDto(list *patients.ListResult, totalCount int) PatientsResponse {
+	data := NewPatientsDto(list.Patients)
 	return PatientsResponse{
 		Data: &data,
-		Meta: &Meta{Count: &list.TotalCount},
+		Meta: &Meta{
+			Count:      &list.MatchingCount,
```
ewollesen, line 587: Would `MatchingCount` be a clearer name?

clinics/clinics_test.go
```diff
@@ -1,12 +1,13 @@
 package clinics_test
 
 import (
+	clinicsTest "github.com/tidepool-org/clinic/clinics/test"
```
ewollesen, line 4: What are you using for a code formatter? Shouldn't this be down below with the other github.com/tidepool entries?

Not a blocker of course.

Roukoswarf (dismissed):

clinics/clinics_test.go
toddkazakov: I'm using IntelliJ which uses go fmt.

PR #184, opened 2025-05-07
batch patient lists for clinic merges

ewollesen:
Our first real attempt to generate a clinic merge report failed
because the patient list exceeded MongoDB's 16 MB Document limit.

Some quick math indicated that more than 1,600 patients could trigger
that limit, and so this code modifies the merge process to list
patients in batches of 1,024 at a time.

The test for this is very slow (due to patient creation), so I welcome
suggestions for improving that (with the understanding that they might
have to come later, this is a high priority backend hero ticket at the
moment.)

SIR-603
BACK-3684 (created after the fact)

### DIFF
--- a/clinics/merge/clinic.go
+++ b/clinics/merge/clinic.go
@@ -344,6 +344,12 @@ func (c *ClinicPlanExecutor) Execute(ctx context.Context, plan ClinicMergePlan) 
 			if err := ppe.Execute(ctx, p, plan.Source, plan.Target); err != nil {
 				return nil, err
 			}
+			if p.SourcePatient != nil {
+				sanitizePatient(p.SourcePatient)
+			}
+			if p.TargetPatient != nil {
+				sanitizePatient(p.TargetPatient)
+			}
 			if err := c.persistPlan(ctx, NewPersistentPlan(planId, planTypePatient, p)); err != nil {
 				return nil, err
 			}
--- a/clinics/merge/clinic.go
+++ b/clinics/merge/clinic.go
@@ -3,18 +3,20 @@ package merge
 import (
 	"context"
 	"fmt"
+	"slices"
+	"time"
+
+	"go.mongodb.org/mongo-driver/bson/primitive"
+	"go.mongodb.org/mongo-driver/mongo"
+	"go.uber.org/fx"
+	"go.uber.org/zap"
+
 	"github.com/tidepool-org/clinic/clinicians"
 	"github.com/tidepool-org/clinic/clinics"
 	"github.com/tidepool-org/clinic/clinics/manager"
 	errs "github.com/tidepool-org/clinic/errors"
 	"github.com/tidepool-org/clinic/patients"
 	"github.com/tidepool-org/clinic/store"
-	"go.mongodb.org/mongo-driver/bson/primitive"
-	"go.mongodb.org/mongo-driver/mongo"
-	"go.uber.org/fx"
-	"go.uber.org/zap"
-	"slices"
-	"time"
 )
 
 const (
@@ -198,25 +200,29 @@ func (m *ClinicMergePlanner) CliniciansMergePlan(ctx context.Context, source, ta
 
 func (m *ClinicMergePlanner) listAllPatients(ctx context.Context, clinic clinics.Clinic) ([]patients.Patient, error) {
 	clinicId := clinic.Id.Hex()
-	limit := 1000000
 
+	filter := patients.Filter{ClinicId: &clinicId}
+	// A randomly chosen patient record (with a summary) measures a little less than 10k
+	// bytes. The limit of a MongoDB Document is 16 MB. That means limit should be ≲
+	// 1600. Let's go with 1024; a nice power of 2.
+	limit := 1024
 	page := store.DefaultPagination().WithLimit(limit)
-	filter := patients.Filter{
-		ClinicId: &clinicId,
-	}
-	result, err := m.patients.List(ctx, &filter, page, nil)
-	if err != nil {
-		return nil, err
-	}
-	if result.TotalCount > limit {
-		return nil, fmt.Errorf("too many patients in clinic")
-	}
+	sort := []*store.Sort{{Attribute: "_id", Ascending: true}}
+	list := []patients.Patient{}
 
-	list := make([]patients.Patient, 0, len(result.Patients))
-	for _, p := range result.Patients {
-		list = append(list, *p)
+	for {
+		result, err := m.patients.List(ctx, &filter, page, sort)
+		if err != nil {
+			return nil, err
+		}
+		for _, p := range result.Patients {
+			list = append(list, *p)
+		}
+		if len(list) == result.TotalCount || len(result.Patients) == 0 {
+			break
+		}
+		page.Offset += len(result.Patients)
 	}
-
 	return list, nil
 }
 
--- a/clinics/merge/clinic.go
+++ b/clinics/merge/clinic.go
@@ -201,8 +201,22 @@ func (m *ClinicMergePlanner) CliniciansMergePlan(ctx context.Context, source, ta
 func (m *ClinicMergePlanner) listAllPatients(ctx context.Context, clinic clinics.Clinic) ([]patients.Patient, error) {
 	clinicId := clinic.Id.Hex()
 
-	filter := patients.Filter{ClinicId: &clinicId, ExcludeSummary: true}
-	limit := 100000 // When ExcludeSummary is set on the filter, this limit should be fine.
+	filter := patients.Filter{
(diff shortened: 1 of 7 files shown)


clinics/merge/clinic.go
```diff
@@ -198,25 +200,29 @@ func (m *ClinicMergePlanner) CliniciansMergePlan(ctx context.Context, source, ta
 
 func (m *ClinicMergePlanner) listAllPatients(ctx context.Context, clinic clinics.Clinic) ([]patients.Patient, error) {
 	clinicId := clinic.Id.Hex()
-	limit := 1000000
 
+	filter := patients.Filter{ClinicId: &clinicId}
+	// A randomly chosen patient record (with a summary) measures a little less than 10k
+	// bytes. The limit of a MongoDB Document is 16 MB. That means limit should be ≲
+	// 1600. Let's go with 1024; a nice power of 2.
```
tjotala, line 207: Suggestion: we may want to add a flag to `patients.List` that tells it to cull the `summary` field from each patient record. That's the bulk of the 10k in each patient record, and not at all needed for this particular query. The actual patient record is on the order of 350 bytes.

In fact, we shouldn't return the `summary` field for non-premium clinics in any case.

toddkazakov: @tjotala I believe the aggregation pipeline will still fail as it first fetches the list of patients, transforms the list, and only then can apply the projection.

toddkazakov (approved):

clinics/merge/clinic.go
ewollesen: LOL, I had the same idea at first @tjotala but then @toddkazakov reminded me of the above. :D

tjotala: I was under the impression that the per-document limit is 16MB and we're clearly not exceeding that. The pipeline output itself is limited to 100MB output for certain stages. In the case of the patient list, I believe the choke point is the $facet stage (`generatePaginationFacetStages`) where each facet has 100MB limit. If the data feeding into the $facet used $unset to clear out `summary` field, it _should_ reduce the pressure against that 100MB limit.

https://www.mongodb.com/docs/manual/reference/operator/aggregation/facet/

tjotala: i.e. the $unset would immediately follow the $match here: https://github.com/tidepool-org/clinic/blob/master/patients/repo.go#L204-L205 - so the documents that are fed into the $facet stage that follows are minimized.

ewollesen: > i.e. the $unset would immediately follow the $match here: https://github.com/tidepool-org/clinic/blob/master/patients/repo.go#L204-L205 - so the documents that are fed into the $facet stage that follows are minimized.

It's totally unclear to me how that would affect the pipeline. The docs for `$unset` indicate that it's an alias for `$project` (so I've no idea why it exists, but ok), and then: 

> Using a $project stage at the beginning or middle of a pipeline to reduce the number of fields passed to subsequent pipeline stages is unlikely to improve performance, as the database performs this optimization automatically.

But here we're not really trying to optimize, but rather stay below certain limits, so... Will it work?

To test locally, I'd need... to find a patient with a summary and duplicate it a few thousand times I guess. I might be able to manage that. I can probably find one on an existing QA cluster and copy it.

tjotala: re: $project vs. $unset - it's inclusion vs. exclusion: $project is "I want to only these fields" whereas $unset is "remove these fields from the projection"

With regard to using $project, my point was simply about the $facet stage used by Patients.Count. The docs say:

> As each stage in a $facet executes, the resulting document is limited to 100 megabytes. Note the [allowDiskUse](https://www.mongodb.com/docs/manual/reference/command/aggregate/#std-label-aggregate-cmd-allowDiskUse) flag doesn't affect the 100 megabyte size limit, since $facet can't spill to disk.

So if the resulting patient list coming out of that $facet stage exceeds 100MB, it'll fail. I believe the $project could be used to reduce the per-document size to fit more.

ewollesen:
OK, so I put 2000 patients in a clinic on my local machine.

Then I put 100 patients in a second clinic.

This PR failed. It turns out 1024 is too large. A value of 128 worked however.

I then tried Tapani's suggestion, and used `$unset: "summary"`, with a limit of 100000, and that also worked.

The upside of this original approach is that it doesn't modify the List code in any way. The downside is that we're still "guessing" about what the limit should be.

But we're guessing either way.

I'm going to go ahead with the List modification, as it should help keep us far from the limits. Thoughts or feedback welcome.

cc: @toddkazakov @tjotala

ewollesen:
Another somewhat closer look at patients with summaries, indicates that 10k is waaay too small an estimate.

```db.patients.find({summary:{$exists:true}}).limit(10000).toArray().map((p) => JSON.stringify(p).length).reduce((s,i) => { if (i > 2000) { s.t+=i ; s.n=s.n+1; }; return s}, {t:0,n:0})
{ t: 159288909, n: 9441 }
```

Which comes to 16872 B per patient on average (when summary exists and is > 2000 to begin with). With that plugged into the equation above, the limit needs to be <= ~ 950, IF we use summaries. But I'll be updating this PR with a flag to unset summaries.

clinics/merge/clinic.go
```diff
@@ -198,25 +200,27 @@ func (m *ClinicMergePlanner) CliniciansMergePlan(ctx context.Context, source, ta
 
 func (m *ClinicMergePlanner) listAllPatients(ctx context.Context, clinic clinics.Clinic) ([]patients.Patient, error) {
 	clinicId := clinic.Id.Hex()
-	limit := 1000000
 
+	filter := patients.Filter{ClinicId: &clinicId, ExcludeSummary: true}
```
tjotala, line 204: Nit: for the purpose of merging, wouldn't we also want `ExcludeDemo: true`?

tjotala:
> Another somewhat closer look at patients with summaries, indicates that 10k is waaay too small an estimate.
> 
> ```
> db.patients.find({summary:{$exists:true}}).limit(10000).toArray().map((p) => JSON.stringify(p).length).reduce((s,i) => { if (i > 2000) { s.t+=i ; s.n=s.n+1; }; return s}, {t:0,n:0})
> { t: 159288909, n: 9441 }
> ```
> 
> Which comes to 16872 B per patient on average (when summary exists and is > 2000 to begin with). With that plugged into the equation above, the limit needs to be <= ~ 950, IF we use summaries. But I'll be updating this PR with a flag to unset summaries.

With the $unset, we could potentially up the limit even higher - though this assumes MongoDB keeps data in BSON (not stringified JSON) during aggregation:

```
db.patients.aggregate([{ $limit: 200000 }, { $unset: 'summary' }, { $replaceWith: { size: { $bsonSize: '$$ROOT' } } }, { $group: { _id: null, count: { $sum: 1 }, total: { $sum: '$size' }, avg: { $avg: '$size' } } }, { $unset: '_id' } ])
[ { count: 200000, total: 70299822, avg: 351.49911 } ]
```

Without the $unset (but counting all records, so don't trust the `avg`) we would exceed the 100MB around 9,000:

```
clinic> db.patients.aggregate([{ $limit: 9000 }, { $replaceWith: { size: { $bsonSize: '$$ROOT' } } }, { $group: { _id: null, count: { $sum: 1 }, total: { $sum: '$size' }, avg: { $avg: '$size' } } }, { $unset: '_id' } ])
[ { count: 9000, total: 94520810, avg: 10502.312222222223 } ]
```

tjotala (dismissed):
LGTM - the nit is exactly that, a nit, and there's no reason to up the limit even though we could do so without blowing the limit. This is a good place to be conservative.

clinics/merge/clinic.go
ewollesen: > re: $project vs. $unset - it's inclusion vs. exclusion: $project is "I want to only these fields" whereas $unset is "remove these fields from the projection"

Right... but `{ $project: { foo: 0 } }` is... $unset. So why have two operators that do the same thing? :)

ewollesen:
> With the $unset,

I've re-run this query today: 
```
db.patients.aggregate([{ $limit: 9000 }, { $replaceWith: { size: { $bsonSize: '$$ROOT' } } }, { $group: { _id: null, count: { $sum: 1 }, total: { $sum: '$size' }, avg: { $avg: '$size' } } }, { $unset: '_id' } ])
[ { count: 9000, total: 38202127, avg: 4244.680777777778 } ]
```

And you can see that it's already benefiting from Alex's new summary updates, making the average summary even smaller > 50% smaller in fact.

clinics/merge/clinic.go
ewollesen: Since I was making more tweaks I went ahead and added this.

toddkazakov (approved):

PR #187, opened 2025-05-21
[BACK-3718] Prevent recently active users from appearing in noData

Roukoswarf:
* Prevent recently active users from appearing in noData
* Document the noData condition's intentions


patients/repo.go
```diff
@@ -1289,11 +1289,17 @@ func (r *repository) TideReport(ctx context.Context, clinicId string, params Tid
...
+		// 1. Have no data within the last 8h
+		// AND either of the following:
+		// 2. Have no data within the cutoff, typically the period length being looked at, subtracted from now
+		// 3. Have a dexcom session, and it is not successfully connected
 		selector := bson.M{
-			"clinicId": clinicObjId,
-			"tags":     bson.M{"$all": tags},
+			"clinicId":                        clinicObjId,
+			"tags":                            bson.M{"$all": tags},
+			"summary.cgmStats.dates.lastData": bson.M{"$lt": time.Now().UTC().Add(-8 * time.Hour)},
```
ewollesen, line 1300: what if LastDataCutoff is say... -10 hours and the user's last data cgmStats is -9 hours (and their dexcom conn is fine)

They wouldn't match this filter... is that the intended behavior? Line 1300 makes me think it's not. Should there be a min((now -8 hours), params.LastDataCutoff)?

ewollesen (approved):

patients/repo.go
```diff
@@ -1289,14 +1289,24 @@ func (r *repository) TideReport(ctx context.Context, clinicId string, params Tid
 	}
 
 	{
+		// This specifically catches users who:
+		// 1. Have no data within the last 8h
+		// AND either of the following:
```
ewollesen, line 1294: This AND seems to conflict with the "$or" operator on line 1301.

I'm not sure what the intention was, but one of these seems to be incorrect.

ewollesen (changes requested):

PR #193, opened 2025-07-14
[BACK-3868] Clinic merge bug fixes

toddkazakov:
- Prevent merging of clinics if the target clinic expects unique MRNs but the source clinic contains patients with duplicates
- Execute clinic merge plans using the transaction context
- Prevent clinic workspace merge if the target clinic uses partial SSO but the source clinic has patients which are not members of the target

### DIFF
--- a/clinics/merge/clinicians.go
+++ b/clinics/merge/clinicians.go
@@ -33,6 +33,8 @@ type ClinicianPlan struct {
 
 	SourceClinicName string `bson:"sourceClinicName"`
 	TargetClinicName string `bson:"targetClinicName"`
+
+	Error *ReportError `bson:"error"`
 }
 
 func (c ClinicianPlan) IsPendingInvite() bool {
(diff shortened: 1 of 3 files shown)


clinics/merge/patients.go
```diff
@@ -254,11 +258,14 @@ func (p *PatientMergePlanner) Plan(ctx context.Context) (PatientPlans, error) {
 }
 
 func (p *PatientMergePlanner) getTargetPatientById(userId string) (*patients.Patient, error) {
-	patient, ok := p.targetByUserId[userId]
-	if !ok || patient == nil {
+	pts := p.targetByAttribute[PatientAttributeUserId][userId]
+	if len(pts) == 0 || pts[0] == nil {
```
ewollesen, line 262: Is there ever a time when you'd get a nil patient? Or is that just a remnant from the previous code that should probably be removed?

If you did have a nil, and it was intentional, what is that supposed to mean?

clinics/merge/clinicians.go
```diff
@@ -138,6 +144,10 @@ func (s *SourceClinicianMergePlanner) Plan(ctx context.Context) (ClinicianPlan,
 		}
 	}
 
+	if plan.ClinicianAction == ClinicianActionMove && len(s.target.MembershipRestrictions) > 0 {
```
ewollesen, line 147: If the clinician has the appropriate email domain, is that then something we should allow?

ewollesen (approved):
Just a couple of small things, nothing blocking.

PR #194, opened 2025-07-24
glycemic range and diagnosis type

ewollesen:
- **add glycemic ranges to the patient record**
- **add diagnosis type to the patient record**


api/mappers.go
```diff
@@ -214,9 +215,180 @@ func NewPatientDto(patient *patients.Patient) PatientV1 {
...
+func newGlycemicRangesDtoWithError(ranges patients.GlycemicRanges) (
+	*GlycemicRangesV1, error) {
+
+	switch ranges.Type {
+	case "custom":
+		custom, err := newGlycemicRangesDtoCustom(ranges.Custom)
+		if err != nil {
+			return nil, err
+		}
+		return &GlycemicRangesV1{Type: "custom", Custom: *custom}, nil
```
lostlevels: Same here `patients.GlycemicRangeTypeCustom`

lostlevels (approved):

api/mappers.go
ewollesen: This has to be `Custom` (found here in api), but sure, I can make that change.

api/mappers.go
```diff
@@ -214,9 +215,180 @@ func NewPatientDto(patient *patients.Patient) PatientV1 {
...
+		return &dtoDiagnosisType
+	default:
+		return nil
+	}
+}
+
+func NewGlycemicRangesDto(ranges patients.GlycemicRanges) *GlycemicRangesV1 {
+	v1Ranges, err := newGlycemicRangesDtoWithError(ranges)
+	if err != nil {
+		return nil
```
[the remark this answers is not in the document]
ewollesen: > I'm guessing we haven't decided on the best way to handle an error during mapping yet?

Not really, no. Anything that's decided is a big change touching dozens of files in dozens of places. I have a feeling I'll be back in this code before long, so I think delaying at this time won't be too dangerous.

In theory, the value we're starting from has been validated already. So the biggest risk is if we change the possible values, and don't properly migrate. We're usually pretty good about migrations.

...and if I'm really trying to convince myself, I remind myself that this function operates just like most of the functions here, except this one actually acknowledges that an error state is possible, whereas a number of the others just straight up ignore the possibility, even though it's there.

PR #203, opened 2025-09-09
[BACK-3970] Add custom tide report categories.

lostlevels:


auth/policy.rego
```diff
@@ -665,6 +665,14 @@ allow {
   clinician_has_read_access
 }
 
+# Allow currently authenticated clinician to get tide report settings
```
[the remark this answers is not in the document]
lostlevels: Removed because not part of updated product spec.

patients/tide.go
```diff
@@ -1,24 +1,41 @@
...
+)
 
 type Tide struct {
 	Config   TideConfig   `json:"config"`
 	Results  TideResults  `json:"results"`
 	Metadata TideMetadata `json:"metadata"`
 }
 
+type TideReportParams struct {
+	Period         *string
```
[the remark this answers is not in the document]
lostlevels: This was the way the previous code was written, but since openapi-codegen has fixed a optional pointer bug, updated to not use pointers and rely on zero values.

patients/repo.go
```diff
@@ -1295,70 +1375,208 @@ func PatientsToTideResult(patientsList []*Patient, period string, exclusions *[]
...
+	SummaryFieldSortOrder int                  // Sort order against SummaryField. < 0 for desc, > 0 for ascending, 0 for no sort.
+	SummaryFieldFilters   []summaryFieldFilter // Filters to filter category against while querying, if any.
+}
+
+// summaryFieldFilter allows defines a query on a summary field
+type summaryFieldFilter struct {
+	SummaryField           string
+	SummaryFieldComparator string // SummaryField comparison operator. This is one of the mongodb query operators such as `$gt`, `$lt`, `$not` etc.
+	// ComparatorOperandValue float64
+	ComparatorOperandExpression any // expression to compare SummaryField against - this may be a simple value like an string, int32, long or an objects.  Because it may take different forms it is of type any
```
ewollesen, line 1393: This seems like a bad idea. It's bypassing the type system, and we're encoding mongodb semantics when we probably don't have to... Though I guess this is kinda the file in which we're bound to mongo anyway, so... meh. 

From the usage, I see two cases: a float, and a combination of 
```
				SummaryFieldComparator:      "$not",
				ComparatorOperandExpression: bson.M{"$gt": .04},
```
My first thought is that "not greater than" is the same as saying "less than or equal to", right? Then we could go back to having a simple float value and not having to use `any` and thereby bypassing the type system?

But I see from a comment in other places that maybe that has to do with the fact that some records might not have this field. I'd strongly suggest finding some other way to handle that. 

I think even:

```
// ComparatorOperandExpression is a mongo filter expression.
//
// Common examples: bson.M{"$not": bson.M{"$gt": .25}} or bson.M{"$gte": 0.7} 
ComparatorOperandExpression bson.M
```
is better/safer than just going with `any`.

If you do that, then you might be able to collapse `SummaryFieldComparator` down with it?

patients/repo.go
```diff
@@ -1295,70 +1375,208 @@ func PatientsToTideResult(patientsList []*Patient, period string, exclusions *[]
...
+		SummaryFieldFilters: []summaryFieldFilter{
+			{
+				SummaryField:                "timeInHighPercent",
+				SummaryFieldComparator:      "$gt",
+				ComparatorOperandExpression: 0.25,
+			}},
+	},
+}
+
+// defaultOrderedCategoryNames is the ORDERED listing of categories by name to
```
ewollesen, line 1531: 👍🏻

patients/repo.go
```diff
@@ -1295,70 +1375,208 @@ func PatientsToTideResult(patientsList []*Patient, period string, exclusions *[]
...
+	"timeInVeryLowPercent",
+	"timeInAnyLowPercent",
+	"dropInTimeInTargetPercent",
+	"timeInTargetPercent",
+	"timeCGMUsePercent",
+	"meetingTargets",
+}
+
+// getCategoriesByNames returns a slice of categories with a CategoryName in names. The order of the returned slice matches the order of the name in names.
+func getCategoriesByNames(names []string) []tideCategory {
```
ewollesen, line 1546:
```suggestion
// categoriesByNames returns a slice of categories whose CategoryName is found in names. The order of the returned slice matches the order of the name in names.
func categoriesByNames(names []string) []tideCategory {
```

https://go.dev/doc/effective_go#Getters

patients/repo.go
```diff
@@ -1295,70 +1375,208 @@ func PatientsToTideResult(patientsList []*Patient, period string, exclusions *[]
...
+	cats := make([]tideCategory, 0, len(names))
+	alreadySeen := make(map[string]bool, len(names))
+	for _, name := range names {
+		if !alreadySeen[name] {
+			alreadySeen[name] = true
+			idx := slices.IndexFunc(availableCategories[:], func(cat tideCategory) bool {
+				return cat.CategoryName == name
+			})
+			if idx > -1 {
```
ewollesen, line 1555: I assume you're ok with silently ignoring a provided name that doesn't match a category? That's something that could come back to bite us later.

patients/repo.go
```diff
@@ -1367,40 +1585,44 @@ func (r *repository) TideReport(ctx context.Context, clinicId string, params Tid
 		Config: TideConfig{
 			ClinicId: clinicId,
 			Filters: TideFilters{
-				TimeInVeryLowPercent:      ">0.01",
-				TimeInAnyLowPercent:       ">0.04",
-				DropInTimeInTargetPercent: "<-0.15",
-				TimeInTargetPercent:       "<0.7",
-				TimeCGMUsePercent:         "<0.7",
+				TimeInVeryLowPercent:      strp(">0.01"),
```
ewollesen, line 1588: Is this because these are no longer marked as required in the API spec? Consider if you might want to use `x-go-type-skip-optional-pointer` in TidepoolApi to keep these as plain strings so we don't have more pointer soup?

Or is it `x-omitzero` now?

https://github.com/oapi-codegen/oapi-codegen/discussions/2027

patients/tide.go
```diff
@@ -2,31 +2,44 @@ package patients
...
+	ExtremeHighGlucoseThreshold float64     `json:"extremeHighGlucoseThreshold"`
 }
 
 type TideFilters struct {
-	DropInTimeInTargetPercent string `json:"dropInTimeInTargetPercent"`
-	TimeCGMUsePercent         string `json:"timeCGMUsePercent"`
-	TimeInAnyLowPercent       string `json:"timeInAnyLowPercent"`
-	TimeInTargetPercent       string `json:"timeInTargetPercent"`
-	TimeInVeryLowPercent      string `json:"timeInVeryLowPercent"`
+	DropInTimeInTargetPercent *string `json:"dropInTimeInTargetPercent,omitempty"`
```
ewollesen, line 35: I would love it if we could avoid useless pointers.

patients/repo.go
```diff
@@ -1295,70 +1375,208 @@ func PatientsToTideResult(patientsList []*Patient, period string, exclusions *[]
...
 
-const TideReportPatientLimit = 50
-const TideReportNoDataPatientLimit = 25
+const TideReportPatientLimit = 100
+const TideReportNoDataPatientLimit = 50
+
+type tideCategory struct {
+	CategoryName          string
+	SummaryField          string
+	SummaryFieldSortOrder int                  // Sort order against SummaryField. < 0 for desc, > 0 for ascending, 0 for no sort.
```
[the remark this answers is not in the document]
lostlevels: Updated. I thought unexported structs don't show though? I've seen a mix in the standard library of actually commenting each field with a complete sentence versus short comments inline.

patients/repo.go
lostlevels: Correct.

PR #204, opened 2025-09-11
[BACK-4024] Refactor service and repository interface from implementation

darinkrauss:
- Refactor service and repository interface from implementation
- Split clinic service into separate service and repository implementations
- Update mock generation for consistency
- https://tidepool.atlassian.net/browse/BACK-4024
- Includes the following approved and merged PRs:
  - https://github.com/tidepool-org/clinic/pull/205
  - https://github.com/tidepool-org/clinic/pull/209

### DIFF
--- a/auth/authorizer_test.go
+++ b/auth/authorizer_test.go
@@ -92,6 +92,19 @@ var _ = Describe("Request Authorizer", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
+		It("allows backend services to fetch clinics by share code /v1/clinics/share_code/acmeclinic", func() {
+			input := map[string]interface{}{
+				"path":   []string{"v1", "clinics", "share_code", "acmeclinic"},
+				"method": "GET",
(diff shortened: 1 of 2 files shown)


darinkrauss:
This refactor was necessary for impending changes for twiist patient count updates. I need to add a dependency to the clinics service for the patients repository.

To update a clinic patient count one needs to count the patients. Currently, we get around this with a small bit of injected code when a patient is created or delete, but with the more complicated rules based upon patient data sources (i.e. twiist) that is no longer practical.

So, the incoming change requires the separate service and repository interfaces for patients. Technically not 100% required for clinics or clinicians, but to keep the design similar (and ease separation of service-level versus repository-level logic) I'd like to keep it as-is.

Hopefully that is okay with you.

toddkazakov (dismissed):

lostlevels (dismissed):

toddkazakov (dismissed):

PR #205, opened 2025-09-16
[BACK-4025/BACK-4026] Exclude patients with twiist data source from patient count

darinkrauss:
- Provide additional patient count breakdown for clinic
- Include patient data source provider and state in patient count calculations
- Recalculate clinic patient account when patient data sources change
- Exclude patients with twiist data source from patient count
- Allow service to force a refresh of a clinic patient count
- Get patient clinic ids
- https://tidepool.atlassian.net/browse/BACK-4025
- https://tidepool.atlassian.net/browse/BACK-4026


clinics/manager/manager.go
```diff
@@ -160,45 +161,23 @@ func (c *manager) FinalizeMerge(ctx context.Context, sourceId, targetId string)
...
 	}
 
-	// Refresh patient count of target clinic
-	return c.refreshPatientCount(ctx, targetId)
+	_ = c.clinicsService.RefreshPatientCount(ctx, targetId) // Ignore any error, already logged
+
+	return nil
 }
 
 func (c *manager) GetClinicPatientCount(ctx context.Context, clinicId string) (*clinics.PatientCount, error) {
```
ewollesen, line 175: The duplication of clinic service methods in the clinic manager seems unnecessary to me. It would only lead to confusion, IMO. You've added a couple of methods to the clinic manager that already exist in the clinic service. Is there some need for that duplication that I'm not seeing?

For background, in my head, the clinic manager existed as a way to sort of get around the patient/clinic circular dependency (I *think* Todd once confirmed this in a Slack convo). It was the place for functionality that required both the patient and the clinic repos. That's nothing formal, just what I had picked up from working on clinic. That's part of why duplicating methods from one to the other strikes me as odd.

I think you wrote a bit about this in Slack, I'll have to dig that thread up...

patients/service/service.go
```diff
@@ -110,7 +113,17 @@ func (s *service) Update(ctx context.Context, update patients.PatientUpdate) (*p
...
 
 	s.logger.Infow("updating patient", "userId", existing.UserId, "clinicId", update.ClinicId)
-	return s.patientsRepo.Update(ctx, update)
+	patient, err := s.patientsRepo.Update(ctx, update)
+	if err != nil {
+		return nil, err
+	}
+
+	// Updates to the demo patient user should not affect the patient count
+	if update.UserId != s.config.ClinicDemoPatientUserId {
```
ewollesen, line 122: RefreshPatientCounts allows for more than a single demo patient, but this code doesn't. Is there a reason for that asymmetry? I would expect this code to check for demo users in a similar fashion to RefreshPatientCounts.

patients/service/service.go
```diff
@@ -233,7 +248,17 @@ func (s *service) UpdateLastUploadReminderTime(ctx context.Context, update *pati
...
 func (s *service) AddProviderConnectionRequest(ctx context.Context, clinicId, userId string, request patients.ConnectionRequest) error {
 	s.logger.Infow("adding provider connection request for user", "clinicId", clinicId, "userId", userId, "provider", request.ProviderName)
-	return s.patientsRepo.AddProviderConnectionRequest(ctx, clinicId, userId, request)
+
+	if err := s.patientsRepo.AddProviderConnectionRequest(ctx, clinicId, userId, request); err != nil {
+		return err
+	}
+
+	// Updates to the demo patient user should not affect the patient count
+	if userId != s.config.ClinicDemoPatientUserId {
```
ewollesen, line 257: Same comment as on the Update method, re how demo patients are determined.

ewollesen (changes requested):

patients/service/service.go
```diff
@@ -78,7 +81,7 @@ func (s *service) Create(ctx context.Context, patient patients.Patient) (*patien
 
 	result, err := s.patientsRepo.Create(ctx, patient)
 
-	s.updateClinicPatientCount(ctx, clinicId)
+	_ = s.clinicsService.RefreshPatientCount(ctx, clinicId) // Ignore any error, already logged
```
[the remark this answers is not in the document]
darinkrauss: There is a possible (though very unlikely) code path where a patient is inserted, but the `Create` call fails. In this case, I'd still want the patient count updated. Worst case there is an unneeded refresh.

ewollesen (approved):

PR #212, opened 2025-12-10
[BACK-4088] Bump go version to 1.25.x

lostlevels:


toddkazakov (dismissed):

darinkrauss (approved):
LGTM, but did you mean to pull in the OpenAPI spec changes and the resulting generated changes?

lostlevels:
> LGTM, but did you mean to pull in the OpenAPI spec changes and the resulting generated changes?

Good catch, this had master's changes but some of those commits ended up getting reverted it seems.

PR #220, opened 2026-02-25
Send generic provider requests using the outbox pattern

toddkazakov:


api/patients.go
```diff
@@ -512,9 +512,11 @@ func (h *Handler) ConnectProvider(ec echo.Context, clinicId ClinicId, patientId
 		return err
 	}
 
+	now := time.Now()
 	request := patients.ConnectionRequest{
-		ProviderName: provider,
-		CreatedTime:  time.Now(),
+		ProviderName:   provider,
```
darinkrauss, line 517: FYI - I consider ProviderType and ProviderName a required pair. Yes, we only have one ProviderType currently, 'oauth', but that may change in the future.

PR #223, opened 2026-02-25
[BACK-4197] Add support for GMI in Note Components.

lostlevels:


clinics/clinics.go
```diff
@@ -214,7 +214,8 @@ type TagsSettings struct {
 }
 
 type FlowsheetSettings struct {
-	Icode bool `bson:"icode,omitempty"`
+	Icode               bool `bson:"icode,omitempty"`
+	SendSeparateGMINote bool `bson:"sendSeparateGMINote,omitempty"`
```
toddkazakov: Instead of adding this to flowsheet settings, you should create a dedicated note settings with `IncludeGMIComponent bool` or `IncludeComponents[] string`. The latter is more flexible but I don't know how likely it is for clinics to request other summary stats to be added to the notes.

spec/redox.v1.yaml
```diff
@@ -1415,6 +1415,21 @@ components:
         - Meta
         - Patient
         - Observations
+    noteComponent:
```
toddkazakov: Is this required? This spec is provided by redox so we should minimize the changes that we do so it is easier to maintain.

toddkazakov (changes requested):

spec/clinic.v1.yaml
```diff
@@ -5327,13 +5327,17 @@ components:
...
           type: boolean
           description: Determine if values should be sent in accorance with ICode2 rounding standards, or if we should send the values at higher precision.
-        sendSeparateGMINote:
+      required:
+        - icode
+    ehrNoteSettings.v1:
+      title: Note Settings
+      type: object
+      properties:
+        IncludeGMI:
```
toddkazakov: This should start with a lower case so it is consistent

toddkazakov (approved):

toddkazakov (changes requested):
Just one small issue

spec/redox.v1.yaml
lostlevels: This was mainly to avoid dealing with some of the anonymous structs generated but I see we have some helpers for that in `clinic-worker`.

clinics/clinics.go
```diff
@@ -219,7 +219,7 @@ type FlowsheetSettings struct {
 }
 
 type NoteSettings struct {
-	IncludeGMI bool `bson:"IncludeGmi,omitempty"`
+	IncludeGmi bool `bson:"IncludeGmi,omitempty"`
```
toddkazakov:
```golang
IncludeGMI bool `bson:"includeGmi,omitempty"`
```

spec/clinic.v1.yaml
```diff
@@ -5333,7 +5333,7 @@ components:
       title: Note Settings
       type: object
       properties:
-        IncludeGMI:
+        includeGmi:
```
toddkazakov:
```
includeGMI
```

toddkazakov (changes requested):
Integration tests are failing

toddkazakov (dismissed):

toddkazakov (approved):

spec/clinic.v1.yaml
```diff
@@ -5378,6 +5389,12 @@ components:
             id: hp97ilzsssprv
         flowsheets:
           $ref: '#/components/schemas/ehrFlowsheetSettings.v1'
+        notes:
+          x-go-type-skip-optional-pointer: true
+          x-omitempty: false
+          x-omitzero: true
+          allOf:
```
toddkazakov, line 5396: I don't think allOf is needed here

toddkazakov (approved):

spec/clinic.v1.yaml
lostlevels: This was pretty much done to allow the optional notes struct [be a non-pointer](https://github.com/tidepool-org/TidepoolApi/blob/master/reference/clinic/models/ehrSettings.v1.yaml#L42) to avoid too many `if foo != nil` checks.

PR #226, opened 2026-02-27
Allow backend services to update and delete clinicians

clintonium-119:
Allow orca to update clinician roles and remove a clinician from a clinic

### DIFF
--- a/auth/policy.rego
+++ b/auth/policy.rego
@@ -280,6 +280,16 @@ allow {
   clinician_has_write_access
 }
 
+# Allow backend service to update or delete a clinician
+# PUT /v1/clinics/:clinicId/clinicians/:clinicianId
+# DELETE /v1/clinics/:clinicId/clinicians/:clinicianId
+allow {
+  allowed_methods := {"PUT", "DELETE"}
+  allowed_methods[input.method]
+  input.path = ["v1", "clinics", _, "clinicians", _]
+  is_backend_service
+}
+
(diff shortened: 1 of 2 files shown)


toddkazakov (approved):

auth/policy.rego
```diff
@@ -280,6 +280,16 @@ allow {
...
 }
 
+# Allow backend service to update or delete a clinician
+# PUT /v1/clinics/:clinicId/clinicians/:clinicianId
+# DELETE /v1/clinics/:clinicId/clinicians/:clinicianId
+allow {
+  allowed_methods := {"PUT", "DELETE"}
+  allowed_methods[input.method]
+  input.path = ["v1", "clinics", _, "clinicians", _]
+  is_backend_service
```
tjotala, line 290: super minor nit: seems like the convention for the other policy clauses is to list the role `is_backend_service` as the first item to highlight that it's a privileged op.

I expect that the order has no functional bearing, so this is purely a stylistic comment.

tjotala (approved):

auth/policy.rego
clintonium-119: @tjotala I was going to do that (and even did so), but then I noticed other entries where it's at the bottom, so I moved it back to the bottom where it seems more consistent with the adjacent rule for the same endpoint above it.

Most of the rules from line 600 onward put it at the end, so I'll leave it there.

PR #227, opened 2026-03-05
fix parsing of patient list float filters

ewollesen:
BACK-4251


toddkazakov (approved):

ewollesen:
Moving this into #259. I tried to use the new stacks, but the changes got to be too much and I had to give up and "start over".

PR #228, opened 2026-03-05
[BACK-4187] Add patient creation metadata

toddkazakov:


api/mappers.go
```diff
@@ -217,9 +217,41 @@ func NewPatientDto(patient *patients.Patient) PatientV1 {
...
+		return nil
+	}
+	meta := &patients.CreationMetadata{}
+	if dto.Integration != nil {
+		switch *dto.Integration {
+		case PatientCreationMetadataV1IntegrationRedox:
+			meta.Integration = patients.IntegrationRedox
+		case PatientCreationMetadataV1IntegrationXealth:
+			meta.Integration = patients.IntegrationXealth
+		}
```
ewollesen, line 250: Is it worth adding a default case the handle some newly added case to log something here? Might help with debugging in the future if we add an EHR provider.

ewollesen, line 250: I think it'd be worth adding a default case here to log that an unhandled integration was received.

ewollesen (dismissed):

api/mappers.go
toddkazakov: The API validation will fail with unknown values. All of the mappers assume the DTOs and models are valid.

PR #229, opened 2026-03-26
[BACK-4315] Allow backend services to send provider connection requests

clintonium-119:
[BACK-4315]
Allow orca to send provider connection requests to patients.

[BACK-4315]: https://tidepool.atlassian.net/browse/BACK-4315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/api/patients.go
+++ b/api/patients.go
@@ -503,8 +503,11 @@ func (h *Handler) ConnectProvider(ec echo.Context, clinicId ClinicId, patientId 
 	ctx := ec.Request().Context()
 
 	authData := auth.GetAuthData(ctx)
-	if err := authData.AssertAuthenticatedUser(); err != nil {
-		return err
+	if authData == nil || authData.SubjectId == "" {
+		return &echo.HTTPError{
+			Code:    http.StatusBadRequest,
+			Message: "expected authenticated user id",
+		}
 	}
 
 	provider, err := NewDataProvider(providerId)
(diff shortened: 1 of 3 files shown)


toddkazakov (dismissed):

clintonium-119:
@toddkazakov Need an additional review - the ConnectProvider handler needed to be updated to allow server tokens.  Same pattern used as `CreatePatientAccount`, but it does leave the `AssertAuthenticatedUser` function as dead code now.

Let me know if I should remove `AssertAuthenticatedUser` now, or if you'd like to keep it there for potential use down the road.

darinkrauss (changes requested):
I'll let Todd approve the PR, but can we get a negative test for this endpoint (to show that unauthenticated users get an error)? To me, that seems much more important than a positive test that the backend service can access it.

clintonium-119:
> I'll let Todd approve the PR, but can we get a negative test for this endpoint (to show that unauthenticated users get an error)? To me, that seems much more important than a positive test that the backend service can access it.

Yep.  Will do.

toddkazakov:
Good catch @darinkrauss, I thought we already had some tests.

toddkazakov (approved):

darinkrauss (approved):
LGTM!

PR #231, opened 2026-04-15
re-introduce patient counts for clinic tags and sites

ewollesen:
The previous implementation ran into unexpected performance issues due
to a query that read all clinics. This case is now detected and
prevented from occurring.

BACK-4274


api/clinicians.go
```diff
@@ -255,11 +255,19 @@ func (h *Handler) listClinics(ec echo.Context, filter clinicians.Filter, page st
 		return err
 	}
 
+	if len(cliniciansList) == 0 {
```
ewollesen: This if statement, and the one a few lines below are the core of the fix for the performance issues.

redox/redox.go
```diff
@@ -317,6 +318,7 @@ func (h *Handler) RescheduleSubscriptionOrdersForPatient(ctx context.Context, pa
 		if len(result) < limit {
 			break
 		}
+		page.Offset += len(result)
```
ewollesen, line 321: This was a small bug fix noticed earlier, that got reverted with the original PR.

clinics/repository/repo.go
```diff
@@ -96,7 +299,9 @@ func (r *repository) List(ctx context.Context, filter *clinics.Filter, paginatio
 		SetLimit(int64(pagination.Limit))
 
 	selector := bson.M{}
-	if len(filter.Ids) > 0 {
+	if filter.Ids != nil && len(filter.Ids) == 0 {
+		return nil, fmt.Errorf("empty list of clinic ids to filter with")
```
toddkazakov: I fear this might blow up if called with an empty list of ids - `$in: []` is a valid mongo query.

api/clinicians.go
toddkazakov: There's no need for this early return

api/clinicians.go
```diff
@@ -255,9 +255,17 @@ func (h *Handler) listClinics(ec echo.Context, filter clinicians.Filter, page st
...
+		return ec.JSON(http.StatusOK, ClinicianClinicRelationshipsV1{})
+	}
+
+	clinicIds := make([]string, len(cliniciansList))
+	for i, clinician := range cliniciansList {
+		clinicIds[i] = clinician.ClinicId.Hex()
+	}
+
+	if len(clinicIds) == 0 {
```
toddkazakov: There's no need for this early return

toddkazakov (approved):

clinics/repository/repo.go
ewollesen: I felt that leaving it in to indicate a programmer error was the better call. While it's a valid query, it's one that doesn't make any sense.

api/clinicians.go
ewollesen: These early returns prevent us from calling clinics list with `ids: []`, which as you noted is a valid mongo query, but is also non-sensical, and is what got us into this mess in the first place. We know there will be no clinic matches, so we shouldn't ever even call clinics list. If we do, it returns EVERY clinic, and then we loop over them all, and throw them all a way. It's just a waste of resources.

clinics/repository/repo.go
ewollesen: This has been removed, in the end I agreed with Todd that it isn't necessary, and is an overcomplication.

api/clinicians.go
ewollesen: These early returns have been removed. Since the guard for passing `clinicIds: []` was removed, we can pass `clinicIds: []` again to basically short-circuit the query, and return a nil result, which in turn removes the need for the early returns.

api/clinicians.go
ewollesen: Removed.

api/clinicians.go
```diff
@@ -255,7 +255,7 @@ func (h *Handler) listClinics(ec echo.Context, filter clinicians.Filter, page st
 		return err
 	}
 
-	var clinicIds []string
+	clinicIds := []string{}
```
ewollesen, line 258: This starts us off with `[]`, so we won't send a `nil` and get a crazy big query.

sites/sites.go
```diff
@@ -17,7 +17,7 @@ import (
 type Site struct {
 	Id       primitive.ObjectID `bson:"id" json:"id"`
 	Name     string             `bson:"name" json:"name"`
-	Patients int                `bson:"-" json:"patients,omitzero"`
+	Patients int                `bson:"patients,omitempty" json:"patients,omitzero"`
```
toddkazakov, line 20: This along with the API mapper changes allows overwriting the counts

integration/patientsites_test.go
```diff
@@ -20,109 +27,317 @@ import (
...
+
+		BeforeEach(func() {
+			_, clinic = th.randomPatientInClinic(th.randomClinic())
+		})
+
+		It("has tags that include the count of patients", func() {
+			clinic, err := th.clinics.Get(th.ctx, clinic.Id.Hex())
+			Expect(err).To(Succeed())
+			if len(clinic.PatientTags) <= 0 {
+				Failf("expected >0 patient tags, got %d", len(clinic.PatientTags))
```
toddkazakov: Use `Expect`

integration/patientsites_test.go
```diff
@@ -20,109 +27,317 @@ import (
...
+			}
+			tag := clinic.PatientTags[0]
+			if tag.Patients != 1 {
+				Failf("expected 1 patient, got %d", tag.Patients)
```
toddkazakov: Use `Expect`

integration/patientsites_test.go
```diff
@@ -20,109 +27,317 @@ import (
...
+		It("has tags that keep their patient count when converted to a site", func() {
+			tag := clinic.PatientTags[0]
+			site, err := th.manager.ConvertPatientTagToSite(th.ctx, clinic.Id.Hex(),
+				tag.Id.Hex())
+			Expect(err).To(Succeed())
+			if site == nil {
+				Fail("expected site to not be nil")
+			}
+			if site.Patients == 0 {
+				Failf("expected site.Patients to be > 0, got %d", site.Patients)
```
toddkazakov: Use `Expect`

integration/patientsites_test.go
```diff
@@ -20,109 +27,317 @@ import (
...
+			}
+			if site.Patients != tag.Patients {
+				Failf("expected site.Patients == tag.Patients, got %d vs %d",
```
toddkazakov: Use `Expect`

integration/patientsites_test.go
```diff
@@ -20,109 +27,317 @@ import (
...
+		clinics:              clinics,
+	}
+}
+
+func (th *testHelper) randomClinic() *clinics.Clinic {
+	randomClinic := clinicsTest.RandomClinic()
+	clinic, err := th.clinics.Create(th.ctx, randomClinic)
+	Expect(err).To(Succeed())
+	if len(clinic.PatientTags) <= 0 {
+		Failf("expected >0 patient tags, got %d", len(clinic.PatientTags))
```
toddkazakov: Use `Expect`

integration/patientsites_test.go
```diff
@@ -20,109 +27,317 @@ import (
...
+	}
+	if len(clinic.Sites) < 2 {
+		Failf("expected >= 2 sites, got %d", len(clinic.Sites))
```
toddkazakov: Use `Expect`

integration/patientsites_test.go
```diff
@@ -20,109 +27,317 @@ import (
...
+				dstSite := th.getSiteFromClinic(clinic, (*patient2.Sites)[0])
+				Expect(srcSite.Patients).To(Equal(1), fmt.Sprintf("srcSite: %+v", srcSite))
+				Expect(dstSite.Patients).To(Equal(1), fmt.Sprintf("dstSite: %+v", dstSite))
+
+				merged, err := th.manager.MergeSite(th.ctx, clinic.Id.Hex(),
+					srcSite.Id.Hex(), dstSite.Id.Hex())
+				Expect(err).To(Succeed())
+
+				Expect(merged).To(Not(BeNil()))
+				Expect(srcSite.Patients + dstSite.Patients).To(Equal(1 + 1))
```
toddkazakov: Should this test whether merged is equal to `srcSite.Patients + dstSite.Patients`?

sites/sites.go
ewollesen: Fixed, I believe. The API has numPatients marked read-only now, and the mappers are adjusted to ignore incoming numPatients values.

toddkazakov (approved):

PR #232, opened 2026-04-29
[BACK-4387] Update TIDE category boundaries to be inclusive rounded values.

lostlevels:
Updates the TIDE report query predicates to query on rounded percents and be inclusive thresholds instead of exclusive. See comments in code for more details.

### DIFF
--- a/patients/repository/repository.go
+++ b/patients/repository/repository.go
@@ -1517,17 +1517,37 @@ type tideCategory struct {
 	//          0 - no sorting
 	//        > 0 - sort in ascending order
 	SummaryFieldSortOrder int
-	// SummaryFieldFilters optionally filter a category while querying.
-	SummaryFieldFilters []summaryFieldFilter
+	// [Predicates] returns an array of mongodb find predicates suitable for input to
+	// the `$and` operator - see [queryPredicates].
+	Predicates queryPredicates
 }
 
-// summaryFieldFilter allows defines a query on a summary field
-type summaryFieldFilter struct {
-	SummaryField string
+// queryPredicates is a function that, given the path prefix of the current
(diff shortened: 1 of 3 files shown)


patients/repository/repository.go
```diff
@@ -1517,142 +1517,260 @@ type tideCategory struct {
...
-			{
-				SummaryField:   "timeCGMUsePercent",
-				QueryPredicate: bson.M{"$lte": 0.7},
-			}},
+		Predicates: func(parentFieldPath string) bson.A {
+			field := parentFieldPath + "." + "timeCGMUsePercent"
+			return bson.A{
+				bson.M{field: bson.M{"$lt": 0.70 - halfAPercent}}, // < because round(0.695) = round(69.5%) = 70% so values less than 69.5% round to < 70%.
+				bson.M{"$expr": bson.M{
+					"$lte": bson.A{
```
ewollesen, line 1630: This should be $lt.

While no test case for cgm use time was provided in the ticket, it might be worth adding one. Or modifying one of the existing ones to show this distinction.

ewollesen (changes requested):

patients/repository/repository.go
lostlevels: Probably a good idea to add tests around all the boundaries I reckon.

lostlevels: This is actually correct. The CGM wear time is one of the few categories that uses an exclusive bounds. Added test for it.

ewollesen: I stand corrected. Do you even need the second rounded criteria for this one? It seems like the `bson.M{field: bson.M{"$lt": 0.70 - halfAPercent}}` should be enough to filter them, no?

ewollesen (approved):

PR #233, opened 2026-04-30
[BACK-4199] Add glycemic ranges to the export url's query string.

lostlevels:


xealth/report.go
```diff
@@ -65,6 +65,17 @@ func GenerateReportUrl(baseUrl string, token string, patient patients.Patient, c
 		query.Add("bgUnits", clinic.PreferredBgUnits)
 	}
 
+	switch patient.GlycemicRanges.Type {
+	case patients.GlycemicRangeTypeCustom:
+		query.Add("glycemicRangeType", string(patients.GlycemicRangeTypeCustom))
+		for _, threshold := range patient.GlycemicRanges.Custom.Thresholds {
+			query.Add("glycemicRangeThresholds", fmt.Sprintf("name,%s,upperBound.value,%f,upperBound.units,%s,inclusive,%v", threshold.Name, threshold.UpperBound.Value, threshold.UpperBound.Units, threshold.Inclusive))
```
ewollesen: Right now these names may include commas, which is probably a mistake on our part, but should be addressed here. Super annoying, I know. I guess even if we don't allow commas, we should probably be escaping this name, for safety.
```suggestion
                        name := &bytes.Buffer{}
                        w = csv.NewWriter(name)
                        w.Write([threshold.Name])
                        w.Flush()
			query.Add("glycemicRangeThresholds", fmt.Sprintf("name,%s,upperBound.value,%f,upperBound.units,%s,inclusive,%t", strings.TrimSpace(name.String()), threshold.UpperBound.Value, threshold.UpperBound.Units, threshold.Inclusive))
```

ewollesen (changes requested):

ewollesen (approved):

PR #236, opened 2026-05-15
migrate pending data sources

ewollesen:
BACK-4414


patients/repository/repository.go
```diff
@@ -883,61 +883,24 @@ func (r *repository) UpdateLastUploadReminderTime(ctx context.Context, update *p
...
 	}
 
 	key := "providerConnectionRequests." + request.ProviderName
-	mongoUpdate["$push"] = bson.M{
-		key: bson.M{
-			"$each": bson.A{request},
-			// Prepend, so the most recent request is stored first
-			"$position": 0,
+	update := bson.M{
+		"$push": bson.M{
```
darinkrauss, line 895: Did you want to update the `updatedTime` field? That seems to have been dropped. Or are we considering the provider connection requests a "supplementary" object that doesn't necessarily imply changing the patient itself?

darinkrauss: Also, any concerns with this array being unbounded? Probably not from a real-world-use perspective.

darinkrauss: Also, I know this hasn't changed, but we are storing the `ProviderName` twice, once as the key and once in the value.

darinkrauss (dismissed):
LGTM!

patients/repository/repository.go
ewollesen: > Or are we considering the provider connection requests a "supplementary" object that doesn't necessarily imply changing the patient itself?

This. Since there's no functional or user visible change to the patient, I'm not touching their updatedTime field.

> Also, any concerns with this array being unbounded? Probably not from a real-world-use perspective.

Agreed. I have concerns, but this code predated my changes, and it's a necessary feature to see the history of connection requests.

> Also, I know this hasn't changed, but we are storing the ProviderName twice, once as the key and once in the value.

Yeah, it annoys me too sometimes, but other times its proven useful to have it in both places.

toddkazakov: @ewollesen Databricks uses the patient's updated time to pull new changes. `updatedTime` ought to be updated when a change to the patient object is made, so those changes can be reflected in all dashboards managed in databricks.

toddkazakov (changes requested):

patients/repository/repository.go
ewollesen: > @ewollesen Databricks uses the patient's updated time to pull new changes. `updatedTime` ought to be updated when a change to the patient object is made, so those changes can be reflected in all dashboards managed in databricks.

Oh you're right. I was thinking of the migration when I wrote this wasn't a material change, not for when providerConnectionRequests are being added/modified. Will fix.

xealth/xealth.go
toddkazakov: This path will still create a `pending` data source, even though it's not a valid value anymore

api/mappers.go
toddkazakov: This block should be removed, now that data sources are marked as read only in the spec

toddkazakov (changes requested):
In Xealth we display the state of the data source (see https://github.com/tidepool-org/clinic/blob/master/xealth/program.go#L62-L67). I believe we should change that logic so we can display similar information, but this time based on the provider connection requests.

toddkazakov (approved):

toddkazakov (dismissed):

PR #237, opened 2026-05-18
Eric revert sites tags count2

ewollesen:
This PR is to be merged in the event that there's an issue with my latest deploy for clinic sites and tags counts and I'm not around to fix it (I have an appt this afternoon that I can't miss).


ewollesen:
All is looking good now, so closing this.

PR #243, opened 2026-07-06
[BACK-4518][BACK-4185] Add patient list export functionality and route.

lostlevels:
Commit for the patient list export functionality. Defines new type, ExportedPatient, since fields are unique to this view of the patient. Adds appropriate mapping and logic over existing fields.


lostlevels:
Ignore @darinkrauss, clicked the wrong button, meant to request @ewollesen .

api/patients.go
```diff
@@ -527,3 +528,36 @@ func (h *Handler) ConnectProvider(ec echo.Context, clinicId ClinicId, patientId
...
+		Period:              params.Period,
+		ExporterClinicianID: authData.SubjectId,
+		WorkspaceID:         clinicId,
+		ReportDate:          time.Now(),
+	}
+	exporter, err := export.NewPatientExport(ctx, h.Clinics, h.Clinicians, h.Patients, filter)
+	if err != nil {
+		return err
+	}
+	disposition := fmt.Sprintf("attachment; filename=patients.csv", time.Now().Unix())
```
ewollesen: This appears to be missing a `%d` or something... maybe:
```suggestion
	disposition := fmt.Sprintf("attachment; filename=patients-%d.csv", time.Now().Unix())
```

export/export.go
```diff
@@ -0,0 +1,294 @@
...
+		"Report Date Time",
+		"Exported By", // name
+		"Exported By", // email
+		"Clinic Name",
+		"Workspace ID",
+		"Days of CGM Data Summarized",
+		"Total Patients",
+	}
+	metadata := []string{
+		e.params.ReportDate.Format("2006-01-02 03:04 PM"),
```
ewollesen: No timezone? I guess not the most important thing, but clinics do have a timezone setting, do they not?

export/export.go
```diff
@@ -0,0 +1,294 @@
...
+	return ds.LatestDataTime.Format("2006-01-02")
+}
+
+func fmtDataSourceStatus(ds *patients.DataSource, now time.Time) string {
+	if ds == nil || ds.State == "" {
+		return "NA"
+	}
+	inactiveCutoff := now.Add(-time.Hour * 24 * 2)
+	expiredCutoff := now.Add(-time.Hour * 24 * 30)
+	if (ds.State == patients.DataSourceStatePending || ds.State == patients.DataSourceStatePendingReconnect) && ds.ExpirationTime != nil && ds.ExpirationTime.Before(expiredCutoff) {
```
ewollesen, line 234: The pending and pending reconnect state are slated to be changed with the release of the device issues dashboard, but I'm guessing this will go out first? If so, I'll have to take a note to fix this when I rebase the DID work onto this.

ewollesen, line 234: This seems strange to me. Why are they inactive when they've been expired for > 30 days? I would think they'd be inactive if they're expired at all?

The spec says:
> *implementation note - this isn’t a status we receive from the backend. This status has to be calculated by counting 30 days after the invite was sent (pending) or resent after disconnect (pendingReconnect)

Which I think implies that the invite expires after 30 days, so any expired invite should be considered inactive?

export/export.go
```diff
@@ -0,0 +1,294 @@
...
+		return "inactive"
```
ewollesen: The spec seems to indicate there's an "expired" state... So why are we calling this "inactive"?

export/export.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	return ds.LatestDataTime.Format("2006-01-02")
+}
+
+func fmtDataSourceStatus(ds *patients.DataSource, now time.Time) string {
+	if ds == nil || ds.State == "" {
+		return "NA"
+	}
+	inactiveCutoff := now.Add(-time.Hour * 24 * 2)
+	expiredCutoff := now.Add(-time.Hour * 24 * 30)
```
ewollesen: Probably better to use https://github.com/tidepool-org/clinic/blob/26c9acdc62c14c3fe8fd8c1e60ac1309a35a52ea/patients/patients.go#L31

export/export.go
```diff
@@ -0,0 +1,294 @@
...
+		return ""
+	}
+	return ds.LatestDataTime.Format("2006-01-02")
+}
+
+func fmtDataSourceStatus(ds *patients.DataSource, now time.Time) string {
+	if ds == nil || ds.State == "" {
+		return "NA"
+	}
+	inactiveCutoff := now.Add(-time.Hour * 24 * 2)
```
ewollesen, line 232: This one has a const defined in upcoming DID work, so... hopefully I'll remember to refactor this when I rebase onto this work.

export/export.go
```diff
@@ -0,0 +1,294 @@
...
+	return gr.Custom.Name
+}
+
+func fmtPatientTags(tagIds *[]string, tagNamesById map[string]string) string {
+	if tagIds == nil || len(*tagIds) == 0 {
+		return ""
+	}
+	var tagNames []string
+	for _, tagId := range *tagIds {
+		if name, ok := tagNamesById[tagId]; ok {
```
ewollesen, line 288: Hmm, this will silently ignore any tags found for which we don't know the name... Is that desired? Maybe better to use "<unknown tag>" or something..? We probably don't want to error out, but just ignoring that we couldn't find a name for the tag seems wrong too.

export/export.go
```diff
@@ -0,0 +1,294 @@
...
+	}
+	return strings.Join(tagNames, ", ")
+}
+
+func fmtBirthdate(birthDate *string) string {
+	if birthDate == nil {
+		return ""
+	}
+	dateFormat := "2006-01-02"
+	t, err := time.Parse(dateFormat, *birthDate)
```
ewollesen:
```suggestion
	t, err := time.Parse(time.DateOnly, *birthDate)
```

export/export.go
```diff
@@ -0,0 +1,294 @@
...
+		return "24 hours"
+	}
+	return fmt.Sprintf("%d days", days)
+}
+
+func fmtDataSourceLastDataDate(ds *patients.DataSource) string {
+	if ds == nil || ds.LatestDataTime == nil {
+		return ""
+	}
+	return ds.LatestDataTime.Format("2006-01-02")
```
ewollesen:
```suggestion
	return ds.LatestDataTime.Format(time.DateOnly)
```

ewollesen (changes requested):
The build fails.

There functions called which are missing definitions.

Have you tested this manually?

lostlevels:
> The build fails.
> 
> There functions called which are missing definitions.
> 
> Have you tested this manually?

@ewollesen Forgot to commit a new file. Tested manually.

export/pointer.go
```diff
@@ -0,0 +1,80 @@
...
+	return fmt.Sprintf("%v", math.RoundToEven(*f*shift)/shift)
+}
+
+func pint(i *int) string {
+	if i == nil {
+		return ""
+	}
+	return ""
+}
+
```
ewollesen, line 28: This doesn't look correct to me. (`pint` always returns `""`)

Since this function is used in your CSV generation, any CSV generated is wrong. You might consider adding some tests of the generated CSV that check for things like this and the number of columns (which _seem_ to be mismatched).

export/export.go
```diff
@@ -0,0 +1,294 @@
...
+
+	clinic                 *clinics.Clinic
+	tagNamesById           map[string]string
+	clinicianNamesByUserId map[string]string
+	exportingClinician     *clinicians.Clinician
+	params                 Params
+	days                   int
+}
+
+func (e *exporter) toCSVRow(p *patients.ExportedPatient) []string {
```
ewollesen: I don't think the number of columns emitted here lines up with the header. I calculate 36 columns from the header, but only 32 from here (based on quick line number subtractions)

ewollesen (changes requested):

export/export.go
lostlevels: typo

export/export.go
lostlevels: typo

export/export.go
```diff
@@ -0,0 +1,324 @@
...
+		pint(p.CgmHoursWithData),
+		ptomgdl(p.CgmAverageGlucose),
+		ppct(p.CgmGmi, 0),
+		ppct(p.CgmStdDev, 0),
+		ppct(p.CgmCV, 0),
+		ppct(p.CgmTimeInLevel2Hypo, 0),
+		ppct(p.CgmTimeInLevel1Hypo, 0),
+		ppct(p.CgmTimeInTarget, 0),
+		ppct(p.CgmTimeInLevel2Hyper, 0),
+		ppct(p.CgmTimeInLevel1Hyper, 0),
```
ewollesen: These should be swapped, to match the headers.

patients/repository/repository.go
```diff
@@ -1092,6 +1092,197 @@ func (r *repository) DeletePatientTagFromClinicPatients(ctx context.Context, cli
...
+	matchStage := bson.D{
+		{"$match", bson.M{
+			"clinicId": clinicObjId,
+		}},
+	}
+
+	cgmPathPrefix := fmt.Sprintf("$summary.cgmStats.periods.%s", period)
+	addCGMFieldsStage := bson.D{
+		{"$addFields", bson.M{
+			"cgmLastDataDate":      "$summary.cgmStats.dates.lastData",
```
ewollesen: This value will never be present in the []ExportedPatient read from the cursor, because patients/export.go:30 defines the tag as `cgmLastData` (not `cgmLastDataDate`).

Your test misses it, because you create an ExportedPatient directly, bypassing the bson marshal/unmarshal.

It looks like bgmLastDataDate has the same issue. Maybe others I missed?

ewollesen (changes requested):

api/patients.go
```diff
@@ -527,3 +528,36 @@ func (h *Handler) ConnectProvider(ec echo.Context, clinicId ClinicId, patientId
...
+		ReportDate:          time.Now(),
+	}
+	exporter, err := export.NewPatientExport(ctx, h.Clinics, h.Clinicians, h.Patients, filter)
+	if err != nil {
+		return err
+	}
+	disposition := fmt.Sprintf("attachment; filename=patients-%d.csv", time.Now().Unix())
+	ec.Response().Header().Set(echo.HeaderContentDisposition, disposition)
+	ec.Response().Header().Set(echo.HeaderContentType, "text/csv")
+	ec.Response().WriteHeader(http.StatusOK)
```
ewollesen: If the `Write` call below fails, you don't want to send OK here.

export/export.go
```diff
@@ -0,0 +1,324 @@
...
+func (e *exporter) Write(ctx context.Context, w io.Writer) error {
+	ps, err := e.patientSvc.ListExportedPatients(ctx, e.params.WorkspaceID, e.params.Period)
+	if err != nil {
+		return err
+	}
+	writer := csv.NewWriter(w)
+	metadataTitles := []string{
+		"Report Date Time",
+		"Exported By", // name
+		"Exported By", // email
```
ewollesen: Is there a good reason not to name these `Exported By (Name)` and `Exported By (Email)` just for clarity? Isn't that what's in PROD-4185?

export/export.go
```diff
@@ -0,0 +1,324 @@
...
+	"time"
+
+	"github.com/tidepool-org/clinic/clinicians"
+	"github.com/tidepool-org/clinic/clinics"
+	"github.com/tidepool-org/clinic/patients"
+	"github.com/tidepool-org/clinic/store"
+)
+
+const (
+	timeFormat = "2006-01-02 03:04 PM"
```
ewollesen, line 19: I'd suggest adding the time zone here, so it's clearer.

export/export.go
```diff
@@ -0,0 +1,324 @@
...
+		fmtDataSourceLastDataDate(p.DexcomDataSource),
+		fmtDataSourceStatus(p.AbbottDataSource, e.params.ReportDate),
+		fmtDataSourceLastDataDate(p.AbbottDataSource),
+		fmtDataSourceStatus(p.TwiistDataSource, e.params.ReportDate),
+		fmtDataSourceLastDataDate(p.TwiistDataSource),
+		ptime(p.CgmLastDataDate, "2006-01-02"),
+		ppct(p.CgmActiveWearTime, 0),
+		pint(p.CgmDaysWithData),
+		pint(p.CgmHoursWithData),
+		ptomgdl(p.CgmAverageGlucose),
```
ewollesen, line 57: Shouldn't these be exported to the clinic's chosen units? The spec says clinic preferred units for CGM Standard Deviation, but not for CGM average glucose, which I assume was a mistake on the author's part..?

export/export.go
```diff
@@ -0,0 +1,324 @@
...
+		ppct(p.CgmGmi, 0),
```
ewollesen: I don't think you want to use ppct here, do you?

```
 db.patients.find({"summary.periods.14d.glucoseManagementIndicator": {$exists:true}},{"summary.periods.14d.glucoseManagementIndicator":true}).limit(5)
[
  {
    _id: ObjectId('61d7217e4faea5d2c9b608bb'),
    summary: { periods: { '14d': { glucoseManagementIndicator: 6.7 } } }
  },
  {
    _id: ObjectId('61d723224faea5d2c9b608bd'),
    summary: { periods: { '14d': { glucoseManagementIndicator: 6.7 } } }
  },
  {
    _id: ObjectId('61e19c644faea5d2c9b608e9'),
    summary: { periods: { '14d': { glucoseManagementIndicator: 7.2 } } }
  },
  {
    _id: ObjectId('61e19c644faea5d2c9b608e8'),
    summary: { periods: { '14d': { glucoseManagementIndicator: 7.2 } } }
  },
  {
    _id: ObjectId('61e19c64e6cc5525b08332fd'),
    summary: { periods: { '14d': { glucoseManagementIndicator: 7.1 } } }
  }
]```

A GMI of 710 isn't typical from my understanding.

Claude is good at finding these things (it found this one). I'd urge you to go over all of these, since your test doesn't pull from a patient with real data (you just manually build a `ExportedPatient` instead) so it can't expose these sorts of things.

File: export/export.go [hunk 19]
```diff
@@ -0,0 +1,324 @@
...
+		ppct(p.CgmStdDev, 0),
+		ppct(p.CgmCV, 0),
+		ppct(p.CgmTimeInLevel2Hypo, 0),
+		ppct(p.CgmTimeInLevel1Hypo, 0),
+		ppct(p.CgmTimeInTarget, 0),
+		ppct(p.CgmTimeInLevel1Hyper, 0),
+		ppct(p.CgmTimeInLevel2Hyper, 0),
+		ptime(p.BgmLastDataDate, "2006-01-02"),
+		ptomgdl(p.BgmAverageGlucose),
```
ewollesen, line 67: Again this should probably be clinic units? The example in BACK-4185 is 105, which certainly looks like a value in mg/dL.

ewollesen (changes requested):

export/export.go
lostlevels: No good reason.

export/export.go
lostlevels: It's always mg/dL.

export/export.go
lostlevels: The average glucose values are in mg/dL

export/export.go
ewollesen: Plus, any tag not found will result in an empty result in the output, like "Foo," or "Foo,,Bar".

export/export.go
```diff
@@ -0,0 +1,339 @@
...
+	return ds.LatestDataTime.Format(time.DateOnly)
+}
+
+func fmtDataSourceStatus(ds *patients.DataSource, now time.Time) string {
+	if ds == nil || ds.State == "" {
+		return "NA"
+	}
+	inactiveCutoff := now.Add(-time.Hour * 24 * 2)
+	expiredCutoff := now.Add(-time.Duration(math.Abs(float64(patients.PendingDataSourceExpirationDuration))))
+	if (ds.State == patients.DataSourceStatePending || ds.State == patients.DataSourceStatePendingReconnect) && ds.ExpirationTime != nil && ds.ExpirationTime.Before(expiredCutoff) {
```
ewollesen: ds.ExpirationTime is the expiration time of the invite. So you should be comparing ds.ExpirationTime.Before(now) to see if its expired, not now - 30 days.

ewollesen (dismissed):
The expiration time calculation is I believe a bug, so should be fixed.

export/export.go
lostlevels: > Plus, any tag not found will result in an empty result in the output, like "Foo," or "Foo,,Bar".

I think we discussed this earlier and agreed having no output to indicate the presence of a tag, even if nameless, was preferable to omitting the tag altogether.

ewollesen (approved):

PR #244, opened 2026-07-07
[BACK-4486] Xealth flowsheets

toddkazakov:


xealth/observation.go
```diff
@@ -0,0 +1,107 @@
...
+		// Statistics are already formatted by the ehr package; Kind says how
+		// to interpret the value string for the FHIR value[x] field.
+		switch s.Kind {
+		case ehr.KindDate:
+			dt := s.Value
+			component.ValueDateTime = &dt
+		case ehr.KindInteger:
+			i, err := strconv.Atoi(s.Value)
+			if err != nil {
+				continue
```
ewollesen, line 62: Nit: This feels like a hard to debug bug waiting to happen. Is it too complicated to wire up a logger here? (or in the other places of this case statement)

ewollesen (dismissed):

PR #245, opened 2026-07-07
[BACK-4185][BACK-4519] Restrict patient export to clinic tiers 0200+.

lostlevels:


toddkazakov (commented):
I suggest moving this logic to the OPA authorization policy

api/patients.go
```diff
@@ -544,6 +544,16 @@ func (h *Handler) ExportPatientList(ec echo.Context, clinicId ClinicId, params E
...
+	clinic, err := h.Clinics.Get(ctx, clinicId)
+	if err != nil {
+		return err
+	}
+	if !clinic.CanExportPatientList() {
+		return &echo.HTTPError{
+			Code:    http.StatusForbidden,
+			Message: "cannot export patient list",
+		}
+	}
```
ewollesen: This feels more like something that should be handled by the export service, rather than the API handler.

I think I'd implement this so that Write returns an error if the clinic doesn't have the appropriate tier.

Then in the ExportPatientList handler, you detect that error, and return your status forbidden, _if_ it's the tier issue error that was returned.

Edit: Or as Todd suggests, putting it in policy.rego works too I guess. :)

toddkazakov: > This feels more like something that should be handled by the export service, rather than the API handler.

The export service deals with data exports (and PDF reports 🤷) and has no knowledge of the clinic service. Let's not overcomplicate this.

auth/authorizer.go
```diff
@@ -75,6 +78,16 @@ func (e *embeddedOpaAuthorizer) Authorize(ctx context.Context, input *openapi3fi
 		in["clinician"] = clinicianStruct.Map()
 	}
 
+	clinic, err := e.getClinicRecord(ctx, input)
```
ewollesen: I'd wrap this such that it's only run for the patients export endpoint. Otherwise this will get run (needlessly) for any endpoint with a clinicID in it.

ewollesen (changes requested):
I think it'd be good to add some tests around the new logic being added here, too.

auth/authorizer_test.go
```diff
@@ -184,6 +192,78 @@ var _ = Describe("Request Authorizer", func() {
...
+				"method": "POST",
+				"auth": map[string]interface{}{
+					"subjectId":    "999999999",
+					"serverAccess": false,
+				},
+			}
+			err := authorizer.EvaluatePolicy(context.Background(), input)
+			Expect(err).To(Equal(auth.ErrUnauthorized))
+		})
+
```
ewollesen, line 266: Cut and paste duplication?

auth/authorizer.go
```diff
@@ -75,6 +78,18 @@ func (e *embeddedOpaAuthorizer) Authorize(ctx context.Context, input *openapi3fi
...
 
+	if input.RequestValidationInput.PathParams[clinicIdPathParameter] != "" && strings.HasSuffix(input.RequestValidationInput.Route.Path, "/export/patients") {
+		clinic, err := e.getClinicRecord(ctx, input)
+		if err != nil {
+			return err
+		}
+		if clinic != nil {
+			clinicStruct := structs.New(*clinic)
+			clinicStruct.TagName = "bson"
+			in["clinic"] = clinicStruct.Map()
```
ewollesen, line 89: Using `in["clinic"] = map[string]interface{}{"tier": clinic.Tier}` here would prevent it from logging the entire clinic map every time this endpoint is called.

Maybe even better would be to re-work with `in["clinic-tier"]`?

auth/authorizer.go
```diff
@@ -134,3 +149,20 @@ func (e *embeddedOpaAuthorizer) getClinicianRecord(ctx context.Context, input *o
...
+
+func (e *embeddedOpaAuthorizer) getClinicRecord(ctx context.Context, input *openapi3filter.AuthenticationInput) (*clinics.Clinic, error) {
+	clinicId := input.RequestValidationInput.PathParams[clinicIdPathParameter]
+	if clinicId == "" {
+		return nil, nil
+	}
+	authData := GetAuthData(input.RequestValidationInput.Request.Context())
+	if authData == nil {
+		return nil, nil
+	}
```
ewollesen: What's this actually doing here?

This doesn't seem like the place to be checking for auth data..?

Maybe this is a cut and paste error from getClinicianRecord, which is actually concerned with that?

ewollesen (changes requested):

auth/authorizer_test.go
lostlevels: yup.

auth/authorizer.go
lostlevels: Indeed.

ewollesen (approved):

PR #246, opened 2026-07-08
 [BACK-4185][BACK-4520] Add auditing of patient export requests.

lostlevels:


patients/service/service.go
```diff
@@ -27,18 +28,20 @@ type service struct {
 	clinicsService   clinics.Service
 	custodialService CustodialService
 	patientsRepo     patients.Repository
+	auditer          audit.AuditEventRecorder
 }
 
 var _ patients.Service = &service{}
 
-func NewService(config *config.Config, repo patients.Repository, clinics clinics.Service, custodialService CustodialService, logger *zap.SugaredLogger, dbClient *mongo.Client) (patients.Service, error) {
+func NewService(config *config.Config, repo patients.Repository, clinics clinics.Service, custodialService CustodialService, auditer audit.AuditEventRecorder, logger *zap.SugaredLogger, dbClient *mongo.Client) (patients.Service, error) {
```
ewollesen, line 36:
```suggestion
func NewService(config *config.Config, repo patients.Repository, clinics clinics.Service, custodialService CustodialService, auditor audit.AuditEventRecorder, logger *zap.SugaredLogger, dbClient *mongo.Client) (patients.Service, error) {
```

patients/service/service.go
```diff
@@ -27,18 +28,20 @@ type service struct {
...
 	return &service{
 		config:           config,
 		dbClient:         dbClient,
 		logger:           logger,
 		clinicsService:   clinics,
 		custodialService: custodialService,
 		patientsRepo:     repo,
+		auditer:          auditer,
```
ewollesen, line 44:
```suggestion
		auditor:          auditor,
```

patients/service/service.go
```diff
@@ -506,8 +509,17 @@ func (s *service) TideReport(ctx context.Context, clinicId string, params patien
...
-func (s *service) ListExportedPatients(ctx context.Context, clinicId, period string) ([]patients.ExportedPatient, error) {
-	return s.patientsRepo.ListExportedPatients(ctx, clinicId, period)
+func (s *service) ListExportedPatients(ctx context.Context, params patients.ExportParams) ([]patients.ExportedPatient, error) {
+	auditEvent := audit.AuditEvent{
+		ClinicianID: params.ExporterClinicianID,
+		ClinicID:    params.WorkspaceID,
+		EventName:   audit.ExportPatientListEvent,
+	}
+	if err := s.auditer.Create(ctx, auditEvent); err != nil {
+		s.logger.Errorf("Error logging patient list export to audit log", "workspaceId", params.WorkspaceID, "error", err)
```
ewollesen, line 519: nit: no need for the word "error" here, it's implied by the logging level, plus the "error" field in the log.
```suggestion
		s.logger.Errorf("logging patient list export to audit log", "workspaceId", params.WorkspaceID, "error", err)
```

patients/export.go
```diff
@@ -4,6 +4,13 @@ import (
 	"time"
 )
 
+type ExportParams struct {
```
ewollesen, line 7: This struct isn't used in this package. Consider moving it to export/export.go and renaming it Params (to avoid repetition https://go.dev/wiki/CodeReviewComments#package-names)

ewollesen (changes requested):
I strongly suggest we use logging instead of auditing for now.

https://tidepoolteam.slack.com/archives/C8R560B6W/p1753450843645129

This is a much bigger subject and shouldn't be solved with a quick one-off change that isn't going to be flexible going forward (for example there's no way to capture a patient id in the current design).

lostlevels:
> I strongly suggest we use logging instead of auditing for now.
> 
> https://tidepoolteam.slack.com/archives/C8R560B6W/p1753450843645129
> 
> This is a much bigger subject and shouldn't be solved with a quick one-off change that isn't going to be flexible going forward (for example there's no way to capture a patient id in the current design).

Agreed, it's also much simpler that in case to add.
Closing this and opened https://github.com/tidepool-org/clinic/pull/248

PR #247, opened 2026-07-14
upgrade go-google.golang.org/grpc to avoid a CVE

ewollesen:
The clinic repo doesn't use GRPC at all, the dependency transitive,
pulled in by echo http framework, but only used when running its
tests. There is no way for this code to be included in a clinic
binary.

That said, Vanta will keep flagging this, and bumping it should be
minimal risk, since it's not something we use anyway.

BACK-4328


ewollesen:
Not moving this forward, since clinic does not use GRPC at all, we'll just mark the vanta issue as not affected.

PR #248, opened 2026-07-17
[BACK-4185][BACK-4520] Audit patient export list requests via logging.

lostlevels:


patients/service/service.go
```diff
@@ -506,8 +506,15 @@ func (s *service) TideReport(ctx context.Context, clinicId string, params patien
...
 
-func (s *service) ListExportedPatients(ctx context.Context, clinicId, period string) ([]patients.ExportedPatient, error) {
-	return s.patientsRepo.ListExportedPatients(ctx, clinicId, period)
+func (s *service) ListExportedPatients(ctx context.Context, params patients.ExportParams) ([]patients.ExportedPatient, error) {
+	ps, err := s.patientsRepo.ListExportedPatients(ctx, params)
+	s.logger.Infow("ListExportedPatients invoked",
+		"clinicId", params.WorkspaceID,
+		"clinicianId", params.ExporterClinicianID,
+		"patientCount", len(ps),
+		"error", err)
```
ewollesen:
```suggestion
    args := []any{
    	"clinicId", params.WorkspaceID,
		"clinicianId", params.ExporterClinicianID,
		"patientCount", len(ps),
	}
	if err !=nil {
	  args = append(args, "error", err)
	}
	s.logger.Infow("ListExportedPatients invoked",
		args...)
```
nit: What do you think of this? It prevents the word "error" from appearing in log messages that don't have an error, which can make it easier when searching in sumologic for actual errors. On the downside it does add some complexity. :(

ewollesen (approved):
Just one comment with food for thought.

ewollesen (approved):

PR #249, opened 2026-07-27
permit backend services to suppress notifications

ewollesen:
The first commit is just cleanup of test names.

The second commit is the (small) fix.

BACK-4174


darinkrauss (approved):
LGTM!

PR #251, opened 2026-08-03
modify backend to support new frontend TIDE dashboard

ewollesen:
BACK-4524


api/mappers.go
```diff
@@ -627,24 +627,27 @@ func NewTideReportParams(params TideReportParams) patients.TideReportParams {
...
-		Sites:          siteIds,
+		Tags:           compactIds(params.Tags),
+		Sites:          compactIds(params.Sites),
 		LastDataCutoff: params.LastDataCutoff,
 		Categories:     categories,
 		ExcludeNoData:  params.ExcludeNoData,
 	}
 }
 
+// compactIds drops empty ids, so that an empty ?tags= or ?sites= value means no filter.
```
darinkrauss, line 640: [nit] Honestly, either of those should be filter everything out. If you don't want the filter don't specify the parameter. This just seems like a cheat to allow a lazy client.

patients/repository/repository.go
```diff
@@ -1853,18 +1855,8 @@ func (r *repository) TideReport(ctx context.Context, clinicId string, params pat
...
-	uniqueSites := make([]string, 0, len(params.Sites))
-	for _, id := range store.ObjectIDSFromStringArray(params.Sites) {
-		if _, ok := seenSites[id]; ok {
-			continue
-		}
-		seenSites[id] = struct{}{}
-		siteIds = append(siteIds, id)               // for mongo filtering
-		uniqueSites = append(uniqueSites, id.Hex()) // for the Tide.Config
-	}
+	tags, uniqueTags := uniqueObjectIds(params.Tags)
```
darinkrauss: `siteIds`, but `tags`?

patients/repository/repository.go
```diff
@@ -1957,17 +1950,17 @@ func (r *repository) TideReport(ctx context.Context, clinicId string, params pat
...
 		//    AND either of the following:
 		//    - Have no data within the cutoff, typically the period length being looked at, subtracted from now
 		//    - Have a dexcom session, and it is not successfully connected
 		selector := bson.M{
-			"_id":      bson.M{"$nin": exclusions},
-			"clinicId": clinicObjId,
-			"tags":     bson.M{"$all": tags},
+			"_id":           bson.M{"$nin": exclusions},
+			"clinicId":      clinicObjId,
+			"dataSources.0": bson.M{"$exists": true},
```
darinkrauss, line 1963: Note that `dataSources` are only for cloud providers. Data uploader from Uploader won't have a `dataSource`. So, as long is this just pertains to cloud providers, then fine, but the parameter `ExcludeNoData` seems more general.

patients/repository/repository.go
```diff
@@ -2233,3 +2228,32 @@ func applySitesFilter(selector bson.M, siteIds []primitive.ObjectID) {
 		selector["sites.id"] = bson.M{"$in": siteIds}
 	}
 }
+
+func applyTagsFilter(selector bson.M, tagIds []primitive.ObjectID) {
+	if len(tagIds) > 0 {
+		selector["tags"] = bson.M{"$all": tagIds}
```
darinkrauss, line 2234: `$all` not `$in`?

darinkrauss (commented):
Not sure if any of these comments require changes. Let me know, if not, and I can approve.

patients/repository/repository.go
ewollesen: > `$all` not `$in`?

Yep, $all for tags, $in for sites. I double checked and that's what product wanted. https://tidepoolteam.slack.com/archives/C053NJ4B2K0/p1785394669785189

api/mappers.go
ewollesen: > [nit] Honestly, either of those should be filter everything out. If you don't want the filter don't specify the parameter. This just seems like a cheat to allow a lazy client.

I could go either way I guess. I don't think anyone is using this in this way, but it felt like Postel's law indicated that we should accept them and throw them away.

patients/repository/repository.go
ewollesen: > Note that `dataSources` are only for cloud providers. Data uploader from Uploader won't have a `dataSource`. So, as long is this just pertains to cloud providers, then fine, but the parameter `ExcludeNoData` seems more general.

That is accurate, this is only meant to be used for patients with a C2C connection, and will be going away before too much longer in favor of the connection issues dashboard (previously known as the device issues dashboard). https://tidepoolteam.slack.com/archives/C053NJ4B2K0/p1785844749517389?thread_ts=1785842034.959809&cid=C053NJ4B2K0

api/mappers.go
darinkrauss: Understood, but I generally don't agree with the common usage of Postel's Law (which Postel never said/wrote as he was specifically referring to TCP, not generalizing, but that's besides the point). While being flexible in accepting inputs is fine, it only works IF ALL parties fully understand where it is flexible and what that flexibility means. 

The `?tags=` could mean "do not filter by tags" (your choice), or "filter by tags, but since there are no tags specified it yields nothing", or "filter by tags, return all that have NO tags specified". The problem here is that there isn't a common understanding of what the flexibility means and that leads to misunderstanding, errors, and ultimately bugs. What seems like the obvious choice to one person may not be obvious to another.

Of course, then the way to solve that vagueness is by documenting it (in the shared documentation, usually the API spec). But, then, with that documentation, aren't we just back to being strict on input?

And, then, all we've done is allow more ways to say the same thing which just means more code to handle multiple possibilities, which never made sense to me. More code equals more potential for bugs and more time maintaining.

Anywho, just a pet peeve of mine, but won't hold this up for it.

patients/repository/repository.go
darinkrauss: Ah, okay. 👍

patients/repository/repository.go
darinkrauss: Okay.

darinkrauss (dismissed):

darinkrauss (approved):
LGTM!

ewollesen:
Moving this into #259. I tried to use the new stacks, but the changes got to be too much and I had to give up and "start over".

PR #256, opened 2026-08-05
Update permissions so only clinic admins, not members, can export.

lostlevels:


auth/policy.rego
ewollesen:
```suggestion
# Allow currently authenticated clinic admin to get patient list export
```

ewollesen (approved):
I think it'd be good to update the comment that I noted, but not blocking.

PR #257, opened 2026-08-06
[BACK-4489] Add bulk patient account creation API.

lostlevels:


patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+	return outputRows, outputHeader, parsedPatients, nil
+}
+
+// CreateCsvPatients takes in a slice of ParsedCsvPatient and creates a patient
+// for each one. It returns a slice of errors corresponding to the patient at
+// each slice index as failure to create a single patient doesn't prevent other
+// patients from being created. It also returns updated CSV output columns as
+// some errors (db-related, system related, etc) can only surface during
+// creation time.
+func CreateCsvPatients(ctx context.Context, patientSvc Service, header []string, patients []ParsedCsvPatient) (outputRows [][]string, errs []error) {
```
ewollesen, line 168: You shouldn't bother returning the errors, none of your code uses them, and it just looks weird to ignore them.

auth/policy.rego
```diff
@@ -833,3 +825,29 @@ allow {
...
 }
+
+# Allow clinic admins to view patients that would be created from bulk account creation.
+# GET /v1/clinics/:clinicId/patients/bulk
+allow {
+  input.method == "GET"
+  input.path = ["v1", "clinics", _, "patients", "bulk"]
+  clinician_has_write_access
+}
+allow {
```
ewollesen, line 836: nit: We usually have comments for the backend services too. Not a big deal to me, but it's been a pretty consistent pattern, so it could be something someone would grep on.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+)
+
+const (
+	// Output CSV is returned w/ original rows along with 2 additional columns
+	OutputColStatus = MaxInputColumns + iota
+	OutputColEmailed
+	NumOutputCols
+)
+
+func ValidateCSVHeader(records [][]string) error {
```
ewollesen, line 53:
```suggestion
func ValidateCsvHeader(records [][]string) error {
```
nit: While I prefer `CSV` > `Csv`[^1], you've used `Csv` everywhere else in this PR.

[^1]: https://go.dev/wiki/CodeReviewComments#initialisms

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+	if len(records) == 0 {
+		return fmt.Errorf("no rows in input records.")
```
ewollesen, line 55:
```suggestion
		return fmt.Errorf("no rows in input records")
```
nit: In this function alone, some errors start with capitals, other have punctuation, some have both. Pick one and stick with it. I personally prefer non-capitalized and no punctuation as it makes them read better when they have context added later.

https://go.dev/wiki/CodeReviewComments#error-strings

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+		return fmt.Errorf(`%w: only have %d columns`, ErrCsvNotEnoughColumns, len(header))
+	}
+	return nil
+}
+
+// ValidCsvPatientValues is a struct of valid values for diagnoses and presets
+// for patients bulk created from a CSV. This is because whether the types are
+// defined in the patients/model layer or api, it will need to be duplicated /
+// mapped from one type to the other but I'm open to other suggestions.
+type ValidCsvPatientValues struct {
```
ewollesen: Ugh, this sucks (edit: the situation sucks, not your solution!).

So what we have here is that you're trying to parse and validate string inputs into either a diagnosis type, or a glycemic ranges preset. Since the OpenAPI  generator built one version of those types, you're trying to re-use them. Unfortunately, that's really awkward, because you're in the patients package, but those types are in the api package, and api imports patients, so patients can't import api without a circular dependency.

I'd rather see something like this:

```go
// ValidDiagnosisTypes is the authoritative source for data model patient diagnosis
// types. This needs to stay in sync with the API DiagnosisTypeV1 enum values defined in the
// clinic client/schema.
var ValidDiagnosisTypes = map[DiagnosisType]struct{}{
      "":              {},
      "gestational":   {},
      "lada":          {},
      "mody":          {},
      "notApplicable": {},
      "other":         {},
      "prediabetes":   {},
      "type1":         {},
      "type2":         {},
      "type3c":        {},
}

func ParseDiagnosisType(raw string) (DiagnosisType, error) {
      dt := DiagnosisType(strings.TrimSpace(raw))
      if _, ok := ValidDiagnosisTypes[dt]; !ok {
              return "", fmt.Errorf("invalid diagnosis type %q", raw)
      }
      return dt, nil
}
```

With that, line 229's check becomes more like:

```go
var diagnosisType *DiagnosisType
if len(record) > ColDiagnosisType {
      dt, err := ParseDiagnosisType(record[ColDiagnosisType])
      if err != nil {
              return nil, err // or accumulate as a row-level validation error
      }
      if dt != "" {
              diagnosisType = &dt
      }
}
```

And I think most importantly, NewPatientsFromColumns no longer needs to accept the `types ValidCsvPatientValues` argument, and the `type ValidCsvpatientValues` goes away.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+	ValidDiagnoses []string
+	ValidPresets   []string
+	DefaultPreset  string
+}
+
+// ParsedCsvPatient represents the potential patient to be created from a Csv
+// row along with a copy of the original input csv row with added information
+// in [Columns]. Patient may be nil if there is an error with the patient,
+// which would be noted in [OutputColStatus] of [Columns]
```
ewollesen, line 80:
```suggestion
// ParsedCsvPatient represents the potential patient to be created from a CSV
// row along with a copy of the original input CSV row with added information
// in [Columns]. Patient may be nil if there is an error with the patient,
// which would be noted in [OutputColStatus] of [Columns]
```
nit: pick one

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+// column defined as the slice index [OutputColStatus] in
+// [ParsedCsvPatient.Columns] in which case [ParsedCsvPatient.Patient] would be
+// empty.
+func ParsePotentialCsvPatients(ctx context.Context, patientSvc Service, userSvc UserService, csvRecords [][]string, clinicId primitive.ObjectID, types ValidCsvPatientValues) (outputRows [][]string, outputHeader []string, parsedPatients []ParsedCsvPatient, err error) {
+	if err := ValidateCSVHeader(csvRecords); err != nil {
+		return nil, nil, nil, err
+	}
+	header := csvRecords[0]
+	outputHeader = make([]string, NumOutputCols)
+	copy(outputHeader, header)
```
ewollesen, line 104:
```suggestion
    // Copy at most MaxInputColumns cells so extra input columns can't leak into the reserved output columns.
    copy(outputHeader[:MaxInputColumns], header)
```
That way, if there are extra columns, we ignore them. For the header, it's not a big deal, because they're overridden in lines 105-106, but line 111 is where it could really matter.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+	outputHeader[OutputColStatus] = "Reason"
+	outputHeader[OutputColEmailed] = "Emailed?"
+	outputRows = append(outputRows, outputHeader)
+
+	for _, record := range csvRecords[1:] {
+		outputRow := make([]string, NumOutputCols)
+		copy(outputRow, record)
```
ewollesen, line 111:
```suggestion
        copy(outputRow[:MaxInputColumns], record)
```
If extra columns are included with the input CSV, they could make their way into the output if they aren't overridden below, like for example if NewPatientForColumns were to error. Consider adding a test case that ensures extra columns in the CSV don't show up in the output.

api/patients.go
```diff
@@ -527,3 +536,50 @@ func (h *Handler) ConnectProvider(ec echo.Context, clinicId ClinicId, patientId
...
+		}
+	}
+	outputRecords, _, _, err := patients.ParsePotentialCsvPatients(ctx, h.Patients, h.Users, records, clinicObjId, validCsvTypes)
+	if err != nil {
+		return err
+	}
+	res := ec.Response()
+	res.Header().Set(echo.HeaderContentType, "text/csv")
+	res.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=bulk-creation-result-%d.csv", time.Now().Unix()))
+	return csv.NewWriter(res.Writer).WriteAll(outputRecords)
```
ewollesen, line 560: This will respond 200, but the spec in TidepoolApi says 201 is returned. I think either is fine, so long as they're consistent.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+	birthDate, err := time.Parse(time.DateOnly, record[ColBirthdate])
+	if err != nil {
+		return nil, fmt.Errorf(`error parsing column "%s" as date: %w`, record[ColBirthdate], err)
+	}
+	mrn := strings.TrimSpace(record[ColMrn])
+	if mrn == "" {
+		return nil, ErrCsvPatientMissingMrn
+	}
+	var email string
+	if len(record) > ColEmail {
```
ewollesen, line 221: By default, csv.NewReader requires that all rows have the same number of columns as the first row, making this check, and the ones like it below useless. Not just because we know they'll be there, but also because a file that doesn't have them will be rejected when the csv.Reader reads the file.

To configure the csv.Reader to accept rows with varying numbers of columns, you need to set the Reader's `FieldsPerRecord` field to a negative value. Or just document that the CSV file has have every row have the same number of columns.

The spec for BACK-4489 says that some of the columns are optional, but doesn't seem to specify if they'll be specified as blank. It's the difference between:

```
one,two,,,
one,two,three,four,five
```
and
```
one,two
one,two,three,four,five
```
So if your inputs are like the first example, you can just drop the `if` guards and all is good. If your inputs are like the second example, you need to set the FieldsPerRecord field to a value < 0 and keep the guards.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+			} else if res != nil && res.MatchingCount > 0 {
+				issues = append(issues, `duplicate MRN`)
+				patientErr = errors.Join(patientErr, ErrCsvPatientDuplicateMrn)
+			}
+			if pstr(patient.Email) != "" {
+				user, err := userSvc.GetUser(*patient.Email)
+				if err != nil && !errors.Is(err, clinicErrs.NotFound) {
+					issues = append(issues, fmt.Sprintf(`system error checking email: %v`, err))
+				} else if user != nil {
+					issues = append(issues, `duplicate email`)
```
ewollesen, line 144: Do you do any checks for multiple MRNs or Emails in the CSV file?

If there were two identical rows in the CSV, they would get past this check.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+			// which there can be multiple.
+			var issues []string
+			page := store.Pagination{Limit: 1, Offset: 0}
+			filter := Filter{
+				ClinicId: strp(string(clinicId.Hex())),
+				Mrn:      patient.Mrn,
+			}
+			res, err := patientSvc.List(ctx, &filter, page, nil)
+			if err != nil {
+				issues = append(issues, fmt.Sprintf(`system error checking mrn: %v`, err))
```
ewollesen, line 134: If we had a database outage while this was running, we'd have a bunch of patients randomly missing, and still return a 200... Is that the intention?

Maybe it's better (it'd certainly be more efficient) to build one query and perform a single list, so that if the one call fails, we fail the entire operation?

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+			} else if res != nil && res.MatchingCount > 0 {
+				issues = append(issues, `duplicate MRN`)
+				patientErr = errors.Join(patientErr, ErrCsvPatientDuplicateMrn)
+			}
+			if pstr(patient.Email) != "" {
+				user, err := userSvc.GetUser(*patient.Email)
```
ewollesen, line 140: Hmm, this one we can't batch up like the List call above. If we're not going to error the process out (on an error other than NotFound that is), we should at least log these errors. When a clinician asks why some of their patients randomly don't show up, it'd be good to be able to tell them there was a temporary network issue or something and if they try again they'll likely show up.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+// creation given a row of text columns.
+func NewPatientFromColumns(record []string, clinicId primitive.ObjectID, types ValidCsvPatientValues) (*Patient, error) {
+	if len(record) < int(NumRequiredColumns) {
+		return nil, fmt.Errorf(`Row has fewer than the minimum required columns: %v`, record)
+	}
+	fullName := strings.TrimSpace(record[ColName])
+	if fullName == "" {
+		return nil, ErrCsvPatientMissingName
+	}
+	if record[ColBirthdate] == "" {
```
ewollesen, line 209: Don't you want to `TrimSpace` this as well?

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+				errs[i] = err
+				status := parsedPatient.Columns[OutputColStatus]
+				if status != "" {
+					status += ", "
+				}
+				status += err.Error()
+				parsedPatient.Columns[OutputColStatus] = status
+				parsedPatient.Columns[OutputColEmailed] = "N"
+			} else {
+				parsedPatient.Columns[OutputColEmailed] = "Y"
```
ewollesen, line 191:
```suggestion
                if parsedPatient.Email != nil && *parsedPatient.Email != "" {
   				    parsedPatient.Columns[OutputColEmailed] = "Y"
   				} else {
   				    parsedPatient.Columns[OutputColEmailed] = "N"
   				}
```
Or something like that? The point being, just cuz they were created successfully, doesn't mean they were emailed.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+	OutputColStatus = MaxInputColumns + iota
+	OutputColEmailed
+	NumOutputCols
+)
+
+func ValidateCSVHeader(records [][]string) error {
+	if len(records) == 0 {
+		return fmt.Errorf("no rows in input records.")
+	}
+	if len(records) == 1 {
```
ewollesen, line 57: This isn't part of the header, so might be better performed outside this function. This function should probably only accept the header itself.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,279 @@
...
+// CSV or otherwise) representing data for a patient in a predefined order and
+// returns the updated csv to be used for output as well as the patients to
+// be created along with their row information in parsedPatients. Individual
+// errors that would prevent a patient from being created but that would NOT
+// stop other patients from being created, if any, are outputed in the "reason"
+// column defined as the slice index [OutputColStatus] in
+// [ParsedCsvPatient.Columns] in which case [ParsedCsvPatient.Patient] would be
+// empty.
+func ParsePotentialCsvPatients(ctx context.Context, patientSvc Service, userSvc UserService, csvRecords [][]string, clinicId primitive.ObjectID, types ValidCsvPatientValues) (outputRows [][]string, outputHeader []string, parsedPatients []ParsedCsvPatient, err error) {
+	if err := ValidateCSVHeader(csvRecords); err != nil {
```
ewollesen, line 99: What if there's no header? Then we'd be silently dropping the first patient. I don't see anything in the spec about the format of the CSV, if the header is required, etc.

ewollesen (changes requested):

patients/bulk_create_patients.go
lostlevels: This was agreed upon as a simplification to assume successful email creation if a user was created (since the actual email happens in clinic-worker / mailer).

patients/bulk_create_patients.go
lostlevels: There is none. For the MVP simplicity I'm assuming there will always be one.

patients/bulk_create_patients.go
ewollesen: I don't think it's such a big ask to have an if-then-else here? It beats having to know that the CSV lies about sending emails.

If there's some reason not to implement this correctly, then let's please add a comment so that future readers have a clue what's going on.

patients/bulk_create_patients.go
ewollesen: > There is none. For the MVP simplicity I'm assuming there will always be one.

That's worth a comment to help debugging in the future I think.

patients/bulk_create_patients.go
lostlevels: I'm fine with that but that does mean someone, me most likely, will have to make a mental note and update the list any time the types change. Now if only the codegen could add these types...

ewollesen: > I'm fine with that but that does mean someone, me most likely, will have to make a mental note and update the list any time the types change. Now if only the codegen could add these types...

I agree that it's a bummer, but it's not a new pattern in clinic. Just about every type in patients/patients.go has a duplicate version generated in the API.

patients/bulk_create_patients.go
```diff
@@ -123,126 +151,147 @@ func ParsePotentialCsvPatients(ctx context.Context, patientSvc Service, userSvc
...
 	}
-	var preset string
+	preset := DefaultGlycemicPreset
 	if len(record) > ColGlycemicPreset {
-		preset = strings.TrimSpace(record[ColGlycemicPreset])
-	}
-	if preset == "" || !slices.Contains(types.ValidPresets, preset) {
-		// Set to Standard if empty or not one of the recognized ones.
-		preset = types.DefaultPreset
+		presetRaw := strings.TrimSpace(record[ColGlycemicPreset])
```
ewollesen: ParseGlycemicRangesPreset also calls TrimSpace, there's no need to call it twice.

api/patients.go
```diff
@@ -527,3 +528,56 @@ func (h *Handler) ConnectProvider(ec echo.Context, clinicId ClinicId, patientId
...
 }
+
+func (h *Handler) ListBulkCreatePatients(ec echo.Context, clinicId ClinicId) error {
+	ctx := ec.Request().Context()
+	clinicObjId, err := primitive.ObjectIDFromHex(clinicId)
+	if err != nil {
+		return err
+	}
+	reader := csv.NewReader(ec.Request().Body)
+	reader.FieldsPerRecord = -1
```
ewollesen: You'll need the same change in the OpenAPI handler that's parsing the CSV body.

```
Name,Birthdate,MRN,Email
Ann Three,2000-01-01,MRN-RAG-003
Bob Four,2000-01-02,MRN-RAG-004,
Cat Five,2000-01-03,MRN-RAG-005,,type1
Dan Six,2000-01-04,MRN-RAG-006,,type2,adaStandard
Eve Seven,2000-01-05,MRN-RAG-007,,,,extra-ignored-column
```

The CSV above is rejected by the OpenAPI generated handler before it ever even hits the real handler.

A clinic integration test seems like a good idea.

api/patients.go
```diff
@@ -527,3 +528,56 @@ func (h *Handler) ConnectProvider(ec echo.Context, clinicId ClinicId, patientId
...
+		}
+	}
+	outputRecords, _, _, err := patients.ParsePotentialCSVPatients(ctx, h.Patients, h.Users, records, clinicObjId)
+	if err != nil {
+		return err
+	}
+	res := ec.Response()
+	res.Header().Set(echo.HeaderContentType, "text/csv")
+	res.Header().Set("Content-Disposition", fmt.Sprintf("attachment; filename=bulk-creation-result-%d.csv", time.Now().Unix()))
+	res.WriteHeader(http.StatusCreated)
```
ewollesen: Your TidepoolApi PR declares that this returns 200, and I think that's the better choice, but either way as long as they're consistent.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,328 @@
...
+	// repeated MRN or email within the CSV itself (e.g., multiple cases of
+	// email "dev@tidepool.org" within the CSV but not associated with an
+	// existing patient.)
+	for _, pp := range parsedPatients {
+		if pp.Patient != nil && mrnCounts[*pp.Patient.Mrn] > 1 {
+			pp.AppendErr(ErrCSVPatientDuplicateMRN)
+		}
+
+		if pp.Patient != nil {
+			if email := pstr(pp.Patient.Email); email != "" && emailCounts[email] > 1 {
```
ewollesen: On line 158 you pass the email through strings.ToLower, but here you don't.

That means your duplicate detection would fail to detect joe@Example.com as a duplicate of joe@example.com.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,328 @@
...
+			diagnosisType = &dt
+		}
+	}
+	preset := DefaultGlycemicPreset
+	if len(record) > ColGlycemicPreset {
+		presetRaw := strings.TrimSpace(record[ColGlycemicPreset])
+		preset = ParseGlycemicRangesPreset(presetRaw, DefaultGlycemicPreset)
+	}
+
+	patient := Patient{
```
ewollesen, line 350: Would it make sense to add InvitedBy to the clinician calling the endpoint, assuming it's a clinician?

If it's a backend service, I guess we'd leave it blank?

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,328 @@
...
+	copy(outputHeader, header)
+	outputHeader[OutputColStatus] = "Reason"
+	outputHeader[OutputColEmailed] = "Emailed?"
+	outputRows = append(outputRows, outputHeader)
+
+	filter := Filter{
+		ClinicId: strp(string(clinicId.Hex())),
+	}
+	page := store.Pagination{Limit: 50_000, Offset: 0}
+	res, err := patientSvc.List(ctx, &filter, page, nil)
```
ewollesen, line 139: I'd suggest setting the ExcludeSummaryExceptFieldsInMergeReports boolean on the filter, so that most of the summaries are left out. Otherwise a big list of patients could overload the MongoDB limits.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,328 @@
...
+	outputRows = append(outputRows, header)
+	for _, parsedPatient := range patients {
+		if parsedPatient.Patient == nil {
+			if errors.Is(parsedPatient.Err(), ErrCSVPatientInvalidEmail) {
+				parsedPatient.Columns[OutputColEmailed] = "N (invalid email)"
+			} else {
+				parsedPatient.Columns[OutputColEmailed] = "N"
+			}
+		} else {
+			_, err := patientSvc.Create(ctx, *parsedPatient.Patient)
```
ewollesen, line 261: I wonder how many patients can be created before the bulk creation request times out?

ewollesen (changes requested):

ewollesen:
Looks like a rebase is also required. :/

patients/bulk_create_patients.go
lostlevels: Updated + updated test.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,332 @@
...
+// ParsePotentialCSVPatients takes an input of slices of string slices (from a
+// CSV or otherwise) representing data for a patient in a predefined order and
+// returns the updated CSV to be used for output as well as the patients to
+// be created along with their row information in parsedPatients. Individual
+// errors that would prevent a patient from being created but that would NOT
+// stop other patients from being created, if any, are outputed in the "reason"
+// column defined as the slice index [OutputColStatus] in
+// [ParsedCSVPatient.Columns] in which case [ParsedCSVPatient.Patient] would be
+// empty. The CSV header is ALWAYS expected.
+func ParsePotentialCSVPatients(ctx context.Context, patientSvc Service, userSvc UserService, csvRecords [][]string, clinicId primitive.ObjectID, invitedBy *string) (outputRows [][]string, outputHeader []string, parsedPatients []*ParsedCSVPatient, err error) {
```
ewollesen, line 120: Most, if not all, of the errors returned by this function will be returned as a 500 error. Most of these errors are validation errors, so a 400 should be returned.

If we return 500, the user will have no indication that their CSV is invalid.

I suggest a test that ensures validation errors return a 400.

patients/bulk_create_patients.go
ewollesen: @lostlevels you haven't responded, but there's still a disconnect here. So I'll repeat it.

You've set FieldsPerRecord to -1, indicating that a ragged CSV is acceptable. However, you've not modified the generated CSV reader in the same way, so ragged CSV files can't be uploaded.

If ragged CSV *IS* acceptable, a change needs to be made to the generated API reader to allow that.

If ragged CSV *IS NOT* acceptable, don't set the FieldsPerRecord here to -1, and remove the guards throughout that check the record to see if the field is present, it will always be present.

Please test this to ensure you have a solution that works how you want it to, or respond explaining more about the situation if I've somehow misunderstood something.

patients/bulk_create_patients.go
```diff
@@ -0,0 +1,332 @@
...
+				outputRow[ColGlycemicPreset] = string(patient.GlycemicRanges.Preset)
+			}
+			if !patient.DiagnosisType.IsZero() {
+				outputRow[ColDiagnosisType] = string(*patient.DiagnosisType)
+			}
+			// No error from column data, now check for issues with mrn / email of
+			// which there can be multiple.
+			mrnCounts[*patient.Mrn]++
+			if mrnCounts[*patient.Mrn] > 1 {
+				parsedPatient.AppendErr(ErrCSVPatientDuplicateMRN)
```
ewollesen, line 178: What if the clinic doesn't have MRN uniqueness set? BACK-4489 seems to assume that MRN uniqueness is a prerequisite, but its never stated. It might be worth double checking with product.

api/patients.go
```diff
@@ -527,3 +528,69 @@ func (h *Handler) ConnectProvider(ec echo.Context, clinicId ClinicId, patientId
...
+	reader := csv.NewReader(ec.Request().Body)
+	reader.FieldsPerRecord = -1
+	records, err := reader.ReadAll()
+	if err != nil {
+		return &echo.HTTPError{
+			Code:    http.StatusBadRequest,
+			Message: fmt.Sprintf(`error reading input csv: %v`, err),
+		}
+	}
+	outputRecords, _, _, err := patients.ParsePotentialCSVPatients(ctx, h.Patients, h.Users, records, clinicObjId, invitedBy)
```
ewollesen: Should there be a test to see if the creation of some of these patients would run into the clinic's patient limit?

Right now, a GET that created 500 patients would succeed, but would the following POST succeed? Or would it run into the patient limit?

What is the intended behavior? It doesn't seem to be addressed in BACK-4489.

auth/policy.rego
ewollesen, line 422: h/t Claude for catching this one

This rule masks your new GET rule, which isn't a big deal, EXCEPT, that this rule works for any clinician, whereas the bulk patient creation is limited to clinic admins, meaning our security isn't what we think it is.

Changing this rule to something like the below might work:

```rego
 allow {
        input.method == "GET"
        input.path = ["v1", "clinics", _, "patients", id]
        id != "bulk"
        clinician_has_read_access
      }
```

cc @toddkazakov because he has good thoughts on URL structure and clobbering like this one.

api/patients.go
ewollesen: @lostlevels this remains an issue.

ewollesen (changes requested):

patients/bulk_create_patients.go
lostlevels: Added test around error checking of ParsePotentialCSVPatients that service calls.

auth/policy.rego
lostlevels: Updated route to `/v1/clinics/:clinicId/bulk/patients` to not step on that route.

ewollesen:
Switching the content-type from text/csv => text/plain does allow the uploading of ragged-column CSV files. Hopefully that doesn't come back to bite us one day. I can imagine a browser automatically detecting the file type as text/csv during upload. Perhaps a content-type attribute can be forced via the `<input>` element to prevent that. I'm not sure if there's a good place to document that. Maybe @christopherasnider has a hidden stash of gotchas.

api/patients.go
lostlevels: Comment in ticket clarifies that only paying clinics will have access to this so no patient limit will exist.

patients/bulk_create_patients.go
lostlevels: Clarified in ticket comment, MRN check done regardless of clinic MRN uniqueness setting.

patients/bulk_create_patients.go
lostlevels: Thoughts on doing this in a transaction? And/Or utilizing mongo's bulkwrite operations? It would still be limited to inter-service calls though.

ewollesen:
> Switching the content-type from text/csv => text/plain does allow the uploading of ragged-column CSV files. Hopefully that doesn't come back to bite us one day. I can imagine a browser automatically detecting the file type as text/csv during upload. Perhaps a content-type attribute can be forced via the `<input>` element to prevent that. I'm not sure if there's a good place to document that. Maybe @christopherasnider has a hidden stash of gotchas.

My latest stance: https://tidepool.atlassian.net/browse/BACK-4489?focusedCommentId=138457

patients/bulk_create_patients.go
ewollesen: > Thoughts on doing this in a transaction? And/Or utilizing mongo's bulkwrite operations? It would still be limited to inter-service calls though.

A transaction doesn't seem helpful to me, at least not at first glance. What do you think you'd gain from it?

Using bulk writes could be helpful, but seems like a ton of work for a tool that won't be used all that frequently. I think adjusting the timeouts for this particular endpoint is a better solution, at least in the short term. If we end up using this a lot or needing it to be faster, we can explore other options then.

ewollesen:
Looks like you'll need another rebase too.

PR #258, opened 2026-08-06
Fix various patient export rounding related issues.

lostlevels:


export/pointer.go
```diff
@@ -34,6 +48,7 @@ func ptomgdl(valMmolL *float64) string {
 	return pfloat(&val, 0)
 }
 
+// ppct takes a number ≤ 1.0 and returns a numbet as a string ≤ 100 with precision digits to the right of the decimal point after banker's rounding.
```
ewollesen:
```suggestion
// ppct takes a number ≤ 1.0 and returns a number as a string ≤ 100 with precision digits to the right of the decimal point after banker's rounding.
```
nit: typo

export/export.go
```diff
@@ -316,9 +305,10 @@ func periodToDays(period string) (days int, err error) {
 	return days, nil
 }
 
+// fmtFloat returns a number as a string with precision digits to the right of the decimal point after banker's rounding.
 func fmtFloat(f float64, precision int) string {
 	shift := math.Pow(10, float64(precision))
-	return fmt.Sprintf("%v", math.RoundToEven(f*shift*100)/shift)
+	return fmt.Sprintf("%v", math.RoundToEven(f)/shift)
```
ewollesen: Are you sure this is what you want?

If you call `fmtFloat(1.234, 2)` it outputs `0.01`. That doesn't seem right.

https://go.dev/play/p/pe56chj98MO

I suggest writing some tests.

ewollesen (changes requested):

patients/repository/repository.go
```diff
@@ -1146,8 +1146,8 @@ func (r *repository) ListExportedPatients(ctx context.Context, params patients.E
 				},
 			},
 			"bgmTotalReadings": bgmPathPrefix + ".totalRecords",
-			"bgmLowEvents":     bgmPathPrefix + ".timeInLowRecords",
-			"bgmHighEvents":    bgmPathPrefix + ".timeInHighRecords",
+			"bgmLowEvents":     bgmPathPrefix + ".timeInVeryLowRecords",  // BGM Low Events are a count of readings below 54 mg/dL hence using the very low threshold count
```
ewollesen, line 1149: wow... yeah, that's... weird. But confirmed with Rea and the docs in Jira.

export/export_test.go
```diff
@@ -14,6 +14,21 @@ import (
 )
 
 var _ = Describe("Export", func() {
+	Describe("Format and conversion functions", func() {
+		DescribeTable("FmtFloat",
+			func(val float64, precision int, expected string) {
```
ewollesen, line 19: I think you've forgotten to actually test anything here.

ewollesen (changes requested):

PR #259, opened 2026-08-11
List Patients endpoint filter fixes and limits

ewollesen:
Previously the spec indicated that floating point parameter values for
List Patients could include unlimited precision. However, this
unlimited precision was not honored by the backend clinic service,
which limited the values to 2 decimal places.

In addition, the filter denied several common use cases, such as `7`
or `10.7` would be rejected.

The spec now allows floating point values to have up to 10 digits in
its mantissa. Integers can be specified, and exponents greater than 9
can be specified.

A good number of tests check this, including one that checks that the
spec matches.

BACK-4552


patients/repository/repository.go
```diff
@@ -1165,7 +1173,34 @@ func (r *repository) UpdateEHRSubscription(ctx context.Context, clinicId, patien
 	return nil
 }
 
-func (r *repository) generateListFilterQuery(filter *patients.Filter) bson.M {
+// ZeroValuesSentinel in a patient-list id filter selects patients with no values assigned.
+const ZeroValuesSentinel = "_"
```
ewollesen, line 1177: This was already being done, but now we're just making it explicit. A matching change is made in TidepoolApi to document what was already happening.

darinkrauss (approved):
LGTM!

PR #261, opened 2026-08-14
upgrade oapi-codegen to v2.8.0

ewollesen:
This brings openapi-3.1 support with it. That gives us enhanced
$ref<->sibling support, and better anyOf support.

It also removes a few workarounds (see xealth_client/fix_types.go).

BACK-4557


ewollesen:
The extra TIDE dashboard releated generation changes are expected.

I prematurely merged a TidepoolApi change ([TidepoolApi #201](https://github.com/tidepool-org/TidepoolApi/pull/201)) before the clinic change was ready to be merged, like a doof. I've looked over it all before, and I'm confident it won't have any effect.

The TidepoolApi changes that were merged early add the ability to specify a sites filter when using the TIDE report, but the backend will just ignore that, so that's no biggie. Our frontend won't be using it until the backend lands.

It removes the requirement for there to be a tags parameter on the TIDE report request, but the frontend will still be sending one (until it gets its matching release). This is the biggest risk, if we think someone other than blip is going to call this, but they still need to be a clinician in the proper clinic to use the endpoint at all, and the performance issues around tags are already gone, so this should be fine. Blip will still be sending the tags parameter until its matching release.

